PR #78, opened 2017-01-25
Extract WP and add reference boards with booter and CLR build

josesimoes:
- Extracted Wire Protocol code from test app
- Add CMake module to support Wire Protocol
- Improve file and function names for better readability
- Extracted functions for HAL interface
- Provide interface functions implementation as weak
- Remove test apps
- Add targets for ST_NUCLEO_F091RC and ST_STM32F4_DISCOVERY
- Isolation of Wire Protocol application process interface
- Rework repo layout
- Major rework on CMake to have simultaneous build of nanoBooter and
nanoCLR images and separate code files and compiler/linker settings
- Improved documentation
- Update board name and RTOS in CMake VS Code config file


src/CLR/Debugger/WireProtocol_Commands.c
```diff
@@ -1,24 +1,35 @@
...
 
 #include "WireProtocol_Commands.h"
 
-// this one is implemented at WireProtocol_Receiver.c
-extern void ReplyToCommand(WP_Message* message, bool fSuccess, bool fCritical, void* ptr, int size);
+//////////////////////////////////////////////////////////////////////
+// helper functions
+
+bool NanoBooter_GetReleaseInfo(ReleaseInfo* releaseInfo)
+{
```
cw2: nanoBooter-specific code in CLR ??

src/CLR/Debugger/WireProtocol_App_Interface.c
```diff
@@ -0,0 +1,112 @@
+//
```
cw2: WireProtocol_App_Interface ? :facepalm:

src/CLR/WireProtocol/WireProtocol.h
```diff
@@ -0,0 +1,164 @@
...
+// This structure is never used, its purpose is to generate a compiler error in case the size of any structure changes.
+struct WP_CompileCheck
+{
+    char buf1[ sizeof(WP_Packet) == 8 * 4 ? 1 : -1 ];
+};
+
+extern uint8_t* marker;
+extern uint16_t lastOutboundMessage;
+extern uint32_t lastPacketSequence;
+extern WP_Message inboundMessage;
```
cw2: Don't expose internals

src/CLR/Debugger/WireProtocol_Commands.c
josesimoes: Don't get fooled by the name. Yes it could be better.

src/CLR/WireProtocol/WireProtocol.h
josesimoes: I always love to learn new stuff. 
Please submit a PR addressing THIS particular issue so I can learn how it should be done.
TIA!

src/CLR/Debugger/WireProtocol_App_Interface.c
josesimoes: The debugger is actually a WIP. This is not doing anything useful right now.
But it's my understanding the it needs to interface with the app layer at some point. Doesn't it?

src/CLR/WireProtocol/WireProtocol.h
cw2: These variables are not part of the public API, they are internal (static) implementation details. If there is a need for them to be shared among multiple source files, they should be placed in a separate "...priv.h" (or similar). Otherwise, they should be declared int the actual source that uses them, WireProtocol_HAL_Interface.c.

The Wire protocol API just defines the contract, it cannot make any assumption about the actual implementation. I might want to use array of inbound messages or use hardware register for packet sequence...

josesimoes: Thanks for the explanations and your thorough review. I guess I'll reach perfection after all... 😉 
(see #79)

PR #89, opened 2017-01-27
Ignore VS dir and user files

cw2:
Obvious fix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,10 @@ cmake-variants.json
 
 # ignore cmaketools user preferences file 
 .vscode/.cmaketools.json
+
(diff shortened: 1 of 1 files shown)


cw2:
Well, I am not sure it is necessary to clutter .gitignore with dozens of lines to ignore files and directories we will never have (?)

josesimoes:
> Well, I am not sure it is necessary to clutter .gitignore with dozens of lines to ignore files and directories we will never have (?)

I agree with you. Maybe just grab from there what is relevant, leaving out the stuff that we are not using/not relevant on/for this project.

cw2:
Already merged as a part of #95.

PR #100, opened 2017-01-31
Add Monitor_Reboot command

josesimoes:
- this is for ChibiOS reference implementation (nanoBooter only)


cw2:
👍

PR #102, opened 2017-01-31
Correct buffer size for ST_NUCLEO_F091RC

### LINKED ISSUE #101: MFDeploy reports NoConnection in ST_NUCLEO_F091RC
Despite this MFDeploy is able to receive the OEM info string!!??
This was working in the test app. Investigate what was changed.

It's still working with ST_STM32F4_DISCOVERY so it must be something related with the transport (serial USB vs serial).

josesimoes:
- fixes #101


cw2:
I have not yet checked all the detail, but isn't the problem really that a packed can be received while previous one has not been processed yet? I think there is a mechanism that prevents that in the orginal NETMF implementation (something along 'set buffer in use' flag)?

josesimoes:
It's in the TX and it's solved. The buffer size got changed when moving from the test app.
This is at the phy layer. There is nothing NETMF layers above can do about it by the time it reaches there.

PR #107, opened 2017-02-03
Add support for ST_NUCLEO144_F746ZG ChibiOS board

josesimoes:
- add files to add support for this board
- add CMake modules to add support for STM32F7 series
- update CMake for ChibiOS to find/support board files located in community overlay and target folders

### DIFF
--- a/CMake/Modules/CHIBIOS_STM32F7xx_GCC_options.cmake
+++ b/CMake/Modules/CHIBIOS_STM32F7xx_GCC_options.cmake
@@ -1,3 +1,8 @@
+#
+# Copyright (c) 2017 The nanoFramework project contributors
+# See LICENSE file in the project root for full license information.
+#
+
 #################################################################
 # WHEN ADDING A NEW SERIES add the appropriate GCC options bellow
 #################################################################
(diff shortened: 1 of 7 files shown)


josesimoes:
😃

piwi1263:
Once this all works for the F7, can #28 be closed too ? @Becafuel 

Oh, and for your information, somebody in the <the-name-not-to-be-mentioned-here> github area managed to get the F7 to speed up 6-7 times, now THAT is interesting. I choose not to x-link that but you look at issue 535 ...

josesimoes:
> Once this all works for the F7, can #28 be closed too ? @Becafuel

That one is related with using the ST Cube MX HAL drivers. So it's not related.

>  in the 'name-not-to-be-mentioned-here' area managed to get the F7 to speed up 6-7 times

We are looking into that so this will be real here too. Out of the box! 😉

cw2:
Regarding F7 vs. F4 speed, IMHO at least half of that is F7 caches (which would be consistent with my results).

It would be nice to know details (i.e. what exactly was measured and how), but I think the odds are low - nothing against the author personally, but he made similar claims/promises before and never released anything (I remember him from Netduino forums, he usually said "...will be open source when ready", which never happened).

Of course there are areas in NETMF that could be improved in terms of speed/performance, but I am a little bit suspecting the author to simply cut things he does not need for his particular use case or make changes that work only with a particular hardware (which is not inherently wrong). But without the information, this is only my personal speculation...

cw2:
We should be able to run ChibiOS testsuite to get credible results.

PR #109, opened 2017-02-03
Add CMake configuration files for target and board level

josesimoes:
- these are parsed by CMake before build to output .h files that are available to be included in code files
- add use example by setting the version numbers (at target level) and the OEM_STRING at board/image level using these configuration files
- update documentation to add instructions on how to use these


cw2:
👍

PR #112, opened 2017-02-04
Minimum required CMake is 3.7

### LINKED ISSUE #82: minimum required CMake is 3.7
- [ ] update root CMakeLists.txt

- [ ] update documentation (build-instructions.md) to mention this

piwi1263:
Fixes #82 

- Changed the minimum required version for CMake to 3.7
- Modified the documentation in the build instructions accordingly.

Signed-off-by: Peter Wessel piwi1263@gmail.com

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@
 # See LICENSE file in the project root for full license information.
 #
 
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.7)
 include(CMakeToolsHelpers OPTIONAL)
 include(ExternalProject)
 
--- a/docs/build-instructions.md
+++ b/docs/build-instructions.md
@@ -16,7 +16,7 @@ The build is based on CMake tool to ease the development in all major platforms.
 
 You'll need:
 - [GNU ARM Embedded Toolchain](https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads)
-- [CMake](https://cmake.org/)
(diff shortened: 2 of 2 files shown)


piwi1263:
Amended the message, but does result in two commits for whatever reason. Edited via kraken but need a pull before it could do a push, so I did now if got a commit extra. One with the message and one without.

josesimoes:
That's probably the way that tool handles the amend operation...
Ideally you would _squash_ those commits into a single one.

cw2:
Hm, that's strange. I don't know this particular client ("kraken"), but the git `amend` command is used specifically to alter the most recent commit without creating an additional one. It is very useful and I do it relatively often (forgot a file etc.).

But, amending doesn't just alter the most recent commit - it replaces it entirely. To git, it will look like a brand new commit. This does not cause any problems as long as it is done in local (unpublished) branch.

In your particular case the branch has been already published and amend has rewritten history - there is a new commit with updated message, different than one in PR branch. This is the reason you cannot sync (push). In order to do that, you have to use 'force' push - there is usually a 'force' or 'discard changes' checkbox in push options, `git push -f` or `git push --force` in command line.

Note: This is the case when rewriting history can cause problem, usually pointed out as one of the biggest git dangers. It's true, if there are people already working on your published branch in their clones/forks, 'force' push will cause them problems. But, because this is only yours pull request branch and nobody else is working on it, it is safe to rewrite it.

Squashing multiple commits into one works in a similar way, the resulting single commit is brand new and the method of syncing the branch is exactly the same as for amend (force push).

PR #113, opened 2017-02-05
Update main CMake for ChibiOS

josesimoes:
- comment code sections for unsupported RTOSes in main CMake

Documentation:
- remove options that are not supported in the current iteration (ChibiOS is the only supported RTOS and HAL)
- correct example command lines


josesimoes:
Aborting this PR...

PR #120, opened 2017-02-15
Fix definitions for VERSION and OEM strings

josesimoes:
- these preprocessor definitions are now defined in project properties
(because for native compilation these are set at target/board level)

### DIFF
--- a/CMake/ChibiOS_target_os.h.in
+++ b/CMake/ChibiOS_target_os.h.in
@@ -11,8 +11,8 @@
 #ifndef _TARGET_OS_H_
 #define _TARGET_OS_H_
 
-#define NANOFRAMEWORK_VERSION_MAJOR @nanoFramework_VERSION_MAJOR@U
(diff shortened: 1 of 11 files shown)


cw2:
What problem does this solve?

Please note Visual C++ project files (*.vcxproj) are temporary, used only for Win32 nanoCLR test - they will be removed during conversion to CMake build.

josesimoes:
The fact that those should be set at platform or board level.
You had to add those there in order to build the debugger project, I guess that's why there is a note there mentioning that those come from platfom_selector.h
Those version numbers are already available in the build, so they should be used from there I guess.
Btw, this is the way the the problem is solved in NETMF equivalent project.

cw2:
Why not use the same mechanism, i.e. to include a header with those symbols?

josesimoes:
That is already happening in native.
This addresses what's required for the win32 debugger project to build

josesimoes:
I've just added a few more changes with renames of the current defines for the version and OEM strings in order to match those already on the CLR.
You can see there that these were already there for sometime just with slightly different names.

PR #127, opened 2017-02-21
Initial work on native nanoCLR build with core files

josesimoes:
- replicated VS solution
- added all required stubbs
- rework existing CMake and add new ones as required
- adjust VS solution to use src files
- plain C WireProtocol implementation currently disabled

*** current nanoBooter has debugger and COM disabled ***


src/CLR/Include/nanoCLR_Runtime.h
```diff
@@ -13,6 +13,18 @@
 #include <nanoCLR_ErrorCodes.h>
 #include <nanoSupport.h>
 
+// definition of WEAK attribute for stub functions
+#if defined(_WIN32)
```
cw2: The correct symbol for VC++ check is `_MSC_VER` (it's possible to use other toolchains to compile for Win32 platform).

src/CLR/Include/nanoHAL.h
```diff
@@ -36,11 +36,26 @@
...
+// because VC++ doesn't support this attribute it's definition end up being an empty one
+#define __nfweak
+
+#elif defined(__GNUC__)
+#define __nfweak __attribute__((weak))
+
+#else
+#error "Unknow platform. Please add definition for weak attribute."
+#endif
+
```
cw2: Avoid duplicate definition - consolidate with definition in `nanoCLR_Runtime.h`, move to a common header (?)

src/CLR/Include/stdafx.h
```diff
@@ -0,0 +1,11 @@
+//
+// Copyright (c) 2017 The nanoFramework project contributors
+// Portions Copyright (c) Microsoft Corporation.  All rights reserved.
+// See LICENSE file in the project root for full license information.
+//
+
+////////////////////////////////////////////////////////////////////////////
+// This is header file is empty on purpose.                               //
+// It has to exist here in order to keep Visual Studio C++ compiler happy //
```
cw2: I don't think so, pre-compiled headers are disabled at the moment for CLR projects.

src/CLR/StartUp/CLRStartup.cpp
```diff
@@ -4,7 +4,586 @@
...
+        {
+            CLR_RT_Assembly* assm;
+            const CLR_RT_Buffer* buffer = (const CLR_RT_Buffer*)it->second;
+            const CLR_RECORD_ASSEMBLY* header = (CLR_RECORD_ASSEMBLY*)&(*buffer)[0];
+
+            // Creates instance of assembly, sets pointer to native functions, links to g_CLR_RT_TypeSystem 
+            NANOCLR_CHECK_HRESULT( LoadAssembly( header, assm ) );
+        }
+#else     
+
```
cw2: Most of the contents does not belong here, it was moved to targets/platform-specific on purpose - e.g. Win32 version uses command line paramenters, ChibiOS version will use something different, ESP32 will use also something different...

src/CLR/WireProtocol/WireProtocol_v2.h
```diff
@@ -157,5 +171,18 @@ struct WP_CompileCheck
...
 };
 
+
+typedef unsigned int       UINT32;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+    
+    UINT32 SUPPORT_ComputeCRC(const void* rgBlock, int nLength, UINT32 crc);
```
cw2: Any specific reason for UINT32 vs. uint32_t? We should choose one and use it consistently (I am considering to replace CLR_XXX types with their std_t counterparts)

src/PAL/BlockStorage/BlockStorageList_stubs.cpp
```diff
@@ -0,0 +1,137 @@
+//
```
cw2: I thought the point of having `__nfweak` implementation was that there would be no need for stubs - `__nfweak` default implementation will be replaced with strong (custom) variant automatically. Or I am missing something here?

targets/CMSIS-OS/ChibiOS/ST_NUCLEO144_F746ZG/nanoCLR/FaultHandlers_GCC.s
```diff
@@ -0,0 +1,59 @@
+@
```
cw2: Why is this needed? Fault/exception handling has to be provided by RTOS.

cw2:
Oh my 😓

cw2:
I need some more time to go through this offline (in my local repo)...

src/CLR/Include/stdafx.h
josesimoes: Better double check that...
1. Look at the original stubs files on the nanoCLR solution, they have the `#include "stdafx.h"`
..\targets\os\win32\nanoCLR\PAL\blockstorageList_stubs.cpp
2. commenting that line causes the build to fail

I'm quite all right with not having this. It was meant as a quick workaround to make it possible to reuse the stubs files on both VS and native.

src/CLR/WireProtocol/WireProtocol_v2.h
josesimoes: This was rename from `WireProtocol.h` to `WireProtocol_v2.h` because the header for the "new" WireProtocol had the same name as the previous one.
And yes, this has been already discussed on that code review. 

I agree on ditching the nanoCLR Types and using the standard ones instead. That probably has to be made on a single pass to replaced them throughout the code base.

src/CLR/Include/stdafx.h
cw2: Ok, to be precise: pre-compiled headers are not used in CLR core projects (libraries). The win32 test application uses pre-compiled headers [only] for its own code..

josesimoes: That's fine and correct but it still won't build without including `stdafx.h`...

targets/CMSIS-OS/ChibiOS/ST_NUCLEO144_F746ZG/nanoCLR/FaultHandlers_GCC.s
josesimoes: You may be right about this...

When looking at the code of these two functions I understood that these were pre-processing and unwinding the stack in order to throw it to a higher level handler that, in turn, would report the exception in a understandable format.

There are indeed fault/exception handling in the RTOS. 
HARD_Breakpoint is being pulled in at line 595(...) of nanoHAL.h
Questions: 
1. Should the RTOS handlers be updated to make them behave as the current handlers to?
2. Is this just a matter of renaming? Because the 'default' ARM name for that vector is HardFault_Handler instead of HARD_Breakpoint

src/PAL/BlockStorage/BlockStorageList_stubs.cpp
josesimoes: You are right _but_ that won't always work. 
In cases where the stubs are required in VS the actual stub implementation (even if empty) has to exist. This is because (as you know) VS doesn't understand stubs.

I'll try to clean up as much stubs as I can adding the declaration as weak.

targets/CMSIS-OS/ChibiOS/ST_NUCLEO144_F746ZG/nanoCLR/FaultHandlers_GCC.s
cw2: Neither 1. nor 2. - HARD_Breakpoint() function has to defined as an appropriate ChibiOS or CMSIS-RTOS call, something like chDbgAssert() or chSysHalt() (have not looked in detail, most likely there is a better function there).

There cannot be any assembly or other code lower than [CMSIS-]RTOS layer in nanoCLR, including boards/targets (with the exception of target/board-specific drivers not supported by RTOS yet).

josesimoes: Both exist. I'll replace this with calls to those.

cw2:
Is there any further work in progress for this?

I think we should make the code a little bit more organized to match the abstraction layers more closely, something like

1. Move nanoHAL* files from `\src\CLR\Include` to `src\HAL\Include`,
2. Move nanoPAL* files from `\src\CLR\Include` to `src\PAL\Include`,

Then we can start adding HAL implementation (ChibiOS-HAL), based on the list of HAL functions required by nanoCLR.

What do you think?

Also, there cannot be any platform-specific code in nanoCLR core source, i.e. CLRStartup.cpp content guarded by PLATFORM_xxx has to be moved to appropriate target implementation.

josesimoes:
Yes  there is! I plan on pushing a commit today.

OK on the move of HAL and PAL to their own folders.

On the HAL implementation specific to each RTOS I think I have a good start on that. It's based on our conversation the other day.

cw2:
Tip: You could have saved some work placing VC++ include paths into `\nf.props` file - it gets included into every CLR vcxproj file, for exactly that purpose (shared things).

cw2:
Great work!

One question: Why are there [identical?] copies of targetHAL files in every board directory? Shouldn't there be only one, because all share the same ChibiOS HAL implementation?

cw2:
Also, I don't still see the reason for the assembly files?

josesimoes:
> Great work!
Thanks 😄 


> One question: Why are there [identical?] copies of targetHAL files in every board directory? Shouldn't there be only one, because all share the same ChibiOS HAL implementation?
Those are chained up to the CMSIS level: board -> ChibiOS -> CMSIS. 
The ideia is:
- provide whatever is available at the highest possible level 
- allow something already defined at an higher level to be easily redefined 

The design is supposed to be explained in the md file at \docs\architecture\HAL.md
(BTW take a close look at that and do comment and suggest changes/improvements)

josesimoes:
> Also, I don't still see the reason for the assembly files?

Just remove all the unused ones and squash again. Let me know how it looks.

cw2:
Tip: You can avoid the need of master back-merges by updating it first and then rebasing your branch on top of (now updated) master.
https://www.atlassian.com/git/tutorials/rewriting-history#git-rebase

josesimoes:
😓  😓  😓  😓  😓  😓  😓  😓

josesimoes:
This should be good now...

PR #128, opened 2017-02-22
Update gitignore with Visual Studio "official" GitHub ignore

josesimoes:
taken from https://github.com/github/gitignore/blob/master/VisualStudio.gitignore


cw2:
I am against unnecessarily big .gitignore files.

I use [TortoiseGit](https://tortoisegit.org/) on Windows and one of it's very nice features is the ability to display [overlay icons](https://tortoisegit.org/about/screenshots/#Overlay_icons_in_explorer) to indicate file/folder git status in Explorer. So, for each file it has to check the entries in .gitignore to determine whether the file is being excluded or not. With this commit, you've added about hundred of match rules, with basically 100% fail rate, because we don't have such files in the repo and in most cases we never will. Why to introduce completely unnecessary overhead (see [wildmatch.c](https://github.com/git/git/blob/master/wildmatch.c))? For repos with large number files it is rather noticeable; and .gitignore contents affects general git commands too.

So, please reconsider... (and just add a single entry for *.VC.db ?)

josesimoes:
Done!

PR #138, opened 2017-02-27
Deleted nanoCLR.VC.db

cw2:
Obvious fix


cw2:
Does the end-of-line characters check work properly in nfbot? It's possible that CR LF is transformed to LF only...

MikroBusNet:
Nothing related to the PR but I am "happy" that someone else than me has issues with DCO checks.
I feel less alone :)

Sorry for the useless comment.

cw2:
@MikroBusNet Someone should have been following the specification and not making the bot too smart. As far as I understand it, I could easily circumvent its DCO check by using the same text in my name and signoff line, as long as it matches the bot will not complain - despite the fact it is not real name (which was the reason for two stage process involving human verification). But I first need to check this with @josesimoes, the process has to be fixed.

PR #139, opened 2017-02-28
Ignore CMake files

cw2:
Signed-off-by: Stanislav Šimíček <cw2@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,10 @@
 # ignore cmake-variants file (will change frequently per user/board setup, use should derive content from cmake-variants.TEMPLATE.json)
 cmake-variants.json
 
+# CMake
+CMakeCache.txt
+CMakeFiles/
+
 # ignore C/C++ launch file (will change frequently per user/board setup, use should derive content from launch.TEMPLATE.json)
 .vscode/launch.json
 


cw2:
Sorry, I don't understand. CMakeCache.txt and CMakeFiles folder are generated during build, they are intermediate artifacts that cannot be included in repository. They has to be added to .gitignore, so they do not appear in the commit made in otherwise unmodified branch:

Current behavior:
1. In fresh master, `git commit` -> no changes detected
2. build STM32F4 discovery image
3. `git commit` -> CMakeCache.txt and CMakeFiles folder detected for addition

Correct/wanted behavior:
1. In fresh master, `git commit` -> no changes detected
2. build STM32F4 Discovery image
3. `git commit` -> **no changes detected**

cw2:
The build was run from <repo>\build\ directory.

josesimoes:
> CMakeCache.txt and CMakeFiles folder are generated during build

That is correct. Both CMakeCache.txt and CMakeFiles folder have to show up inside the build folder (which is already ignored) so there should be no need for a particular ignore for those. Could you please check if this is what you are seeing in your setup?

> The build was run from <repo>\build\ directory; following the second example instructions.

That should make all the build artefacts to go into the build folder... Are you using the full command line up to the double dots `..` at the end?

cw2:
It had to be a problem with one of my previous builds - after using fresh repos, CMake files are generated only in build directory.

josesimoes:
Glad to know that. 
I've found that when something is behaving strange with the build and it was working just before that wiping the build folder and start over is usually all that it takes.

cw2:
Those files were generated in \targets\CMSIS-OS\ChibiOS\ST_STM32F4_DISCOVERY directory, so wiping build did not help (I do that almost always, just to be sure).

cw2:
I probably run cmake from there once...

PR #140, opened 2017-03-01
Add support to allow use of hardware FPU

### LINKED ISSUE #121: Use of CORTEX_USE_FPU TRUE hangs MCU
In chconf.h, setting CORTEX_USE_FPU to TRUE compiles fine but hangs the board when flashed on a STM32F427 MCU using ChibiOS.

Please see [this thread](https://nanoframework.slack.com/conversation/chibios/p1487075678000106) for more detailed informations if needed.

josesimoes:
- add new parameter to main CMake: USE_FPU
- update CMake GCC option files for F4 and F series
- update CMake for reference boards where applicable
- update documentation accordingly
- update (and clean up) cmake-variants template file

Fixes #121


josesimoes:
> Was this tested? I don't think the values are correct

Yes sir. Obviously this was tested for all three reference boards before the commit was made.
On the previous version the -mfloat-abi=softfp was being used and it was running fine. That's probably because the CORTEX_USE_FPU was guarding the configuration of the floating point this causing that no actual FP call was used...
To understand this: are you suggesting that soft should be used for NO FPU. Is this correct?

PR #145, opened 2017-03-02
Implement HAL_Time_TicksToTimeMilliSec

josesimoes:
- rename function to HAL_Time_TicksToTimeMilliSec to clarify output
- changed the output from 'ticks' (100ns intervals) to milliseconds to follow CMSIS API and because having a time base granularity of 1 millisecond is more than enough (100ns intervals is really too much for an embedded system) this also allows an extra efficiency of using 32bits vars and math for this instead of 64bits
- update code accordingly where required
- add ChibiOS implementation


cw2:
Sorry, but I don't understand the reason for change and what is the difference. 100ns ticks has nothing to do with system timer granularity.

josesimoes:
Time base (native target) is provided by CMSIS osKernelSysTick() and returns an uint32_t.
In CMSIS the timer frequency setting for sysTick is specified in milliseconds (not nanoseconds).
Don't see a reason why nanoFramework needs a time base with a resolution higher than 1 millisecond.

If we are using this as the time base and the value is 32 bits I don't see a point on artificially 'extending' this to 64 bits just because that was used in NETMF (and in NETMF it's there with 64bits probably because that's what is used in [FILETIME](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx) ).

My point being: 
- for it's intended use a resolution of 1 millisecond should be more than enough
- using 32 bits is surely more efficient and resource wise (when compared with 64 bits)

cw2:
https://github.com/ARM-software/CMSIS/issues/44

cw2:
Ok, for timeous (which is the actual use), milliseconds should be enough. Use osKernelSysTick() directly
```
auto tick = osKernelSysTick();                      // get start value of the Kernel system tick
do {                                           // poll device busy status for 100 microseconds
...
} while ((osKernelSysTick() - tick) < osKernelSysTickMicroSec(100));  
```
or wrap the timeout call similarly to ChibiOS' `chTimeElapsedSince(...)`.

josesimoes:
> ARM-software/CMSIS#44 

Nothing against that. And I do understand it perfectly. 
But... here at nanoFramework ... as you've just pointed out this is only used for timeouts so relying on osKernelSysTick() _and_ having it as a uint32_t is quite enough.

josesimoes:
> Ok, for timeouts ...........

This you showing an usage example or you are suggesting that this code block should be replacing something?

cw2:
Only usage example (how to implement a timeout check).

src/CLR/WireProtocol/WireProtocol.cpp
```diff
@@ -273,7 +273,7 @@ bool WP_Message::Process()
                 // If the time between consecutive payload bytes exceeds the timeout threshold then assume that
                 // the rest of the payload is not coming. Reinitialize to synch on the next header.
 
-                // FIXME: if(HAL_Time_TicksToTime(curTicks - m_payloadTicks) < (UINT64)c_PayloadTimeout)
+                // FIXME: if(HAL_Time_TicksToTimeMicroSec(curTicks - m_payloadTicks) < (UINT64)c_PayloadTimeout)
```
josesimoes: Done! And I've took the opportunity to uncomment the code lines handling the timeout.

josesimoes:
This calls for a another adjustment. HAL_Time_CurrentTicks() is a proxy to osKernelSysTick which is uint32_t, so it should be scaled to uint32_t.
Are we in agreement on this?

targets/os/win32/Include/targetHAL_Time.h
```diff
@@ -22,15 +22,15 @@
 //
 //#define TIME_ZONE_OFFSET    ((INT64)Time_GetTimeZoneOffset() * 600000000)
 //
-//INT64 HAL_Time_TicksToTime(UINT64 Ticks);
-//UINT64 HAL_Time_TicksToTime();
+
+
 UINT64 HAL_Time_CurrentTicks();
 #define HAL_Time_CurrentTicks           HAL_Time_CurrentTicks
```
josesimoes: Fixed.

targets/os/win32/nanoCLR/Time.cpp
```diff
@@ -22,12 +22,13 @@ UINT64 HAL_Time_CurrentTicks()
     return 0; // UNDONE: FIXME: EmulatorNative::GetITimeDriver()->CurrentTicks();
 }
 
-INT64 HAL_Time_TicksToTime( UINT64 Ticks )
+INT32 HAL_Time_TicksToTimeMicroSec( UINT32 Ticks )
 {
     _ASSERTE(Ticks <= 0x7FFFFFFFFFFFFFFF);
```
josesimoes: Value scaled.

cw2:
No, use osKernelSysTick() instead (either rename directly, or remove the function declaration and `#define HAL_Time_CurrentTicks osKernelSysTick`).

cw2:
But don't forget to find all occurrences and fix possible calculations...

josesimoes:
Not seeing why the return type should remain as UINT64...

src/CLR/Core/GarbageCollector.cpp
```diff
@@ -171,7 +171,7 @@ CLR_UINT32 CLR_RT_GarbageCollector::ExecuteGarbageCollection()
 #if defined(NANOCLR_TRACE_MEMORY_STATS)
     if(s_CLR_RT_fTrace_MemoryStats >= c_CLR_RT_Trace_Info)
     {
-        int milliSec = ((int)::HAL_Time_TicksToTime( HAL_Time_CurrentTicks() - stats_start ) + TIME_CONVERSION__TICKUNITS - 1) / TIME_CONVERSION__TICKUNITS;
+        int milliSec = (int)::HAL_Time_TicksToTimeMicroSec( HAL_Time_CurrentTicks() - stats_start );
```
josesimoes: Corrected calculation to show milliseconds. Should remain that way in order to show a meaningful value.

cw2:
??? Just leave it to what osKernelSysTick() has (that is exactly what #define will do). After it is changed to 64 bits, we will decide what to do - but better fix compiler errors later, than introduce nasty casting/wrapping bugs now.

src/CLR/Core/GarbageCollector.cpp
```diff
@@ -171,7 +171,9 @@ CLR_UINT32 CLR_RT_GarbageCollector::ExecuteGarbageCollection()
 #if defined(NANOCLR_TRACE_MEMORY_STATS)
     if(s_CLR_RT_fTrace_MemoryStats >= c_CLR_RT_Trace_Info)
     {
-        int milliSec = (int)::HAL_Time_TicksToTimeMicroSec( HAL_Time_CurrentTicks() - stats_start );
+        // convert to milliseconds to show a more meaningfull value 
+        // (because of the typical operation time showing it in microseconds could lead to a huge number...)
+        int milliSec = (::HAL_Time_TicksToTimeMicroSec( HAL_Time_CurrentTicks() - stats_start )) * 1000;
```
cw2: Please use a named constant or conversion macro instead of a magic value (*1000). Much easier to track down next time, when the time unit changes :)

targets/CMSIS-OS/ChibiOS/nanoCLR/targetHAL_Time.c
```diff
@@ -13,8 +11,8 @@ typedef unsigned int        UINT32;
 typedef signed   int        INT32;
 //////////////////////////////////////////
 
-// Converts ticks (CMSIS sysTicks) in miliseconds
+// Converts ticks (CMSIS sysTicks) to microseconds
 INT32 HAL_Time_TicksToTimeMicroSec(UINT32 ticks) {
     // T[s] = 1 / f[Hz] 
-    return (ticks / (osKernelSysTickFrequency * NANOHAL_TIME_CONVERSION_MILISECONDS));
+    return (ticks / (osKernelSysTickFrequency * NANOHAL_TIME_CONVERSION_MICROSECONDS));
```
cw2: Was this tested? For STM32F4-Discovery
```
 #define CH_CFG_ST_FREQUENCY                 10000
```
the multiplication overflows.

josesimoes: Glad you spot this... This made me realize that we don't need micro seconds calculation for timeouts because the sys clock is... milliseconds!

targets/CMSIS-OS/ChibiOS/nanoCLR/targetHAL_Time.c
```diff
@@ -11,8 +11,8 @@ typedef unsigned int        UINT32;
...
 //////////////////////////////////////////
 
-// Converts ticks (CMSIS sysTicks) to microseconds
-INT32 HAL_Time_TicksToTimeMicroSec(UINT32 ticks) {
-    // T[s] = 1 / f[Hz] 
-    return (ticks / (osKernelSysTickFrequency * NANOHAL_TIME_CONVERSION_MICROSECONDS));
+// Converts ticks (CMSIS sysTicks) to milliseconds
+INT32 HAL_Time_TicksToTimeMilliSec(UINT32 ticks) {
+    
+    return ticks * (NANOHAL_TIME_CONVERSION_MICRO_TO_SECONDS / osKernelSysTickFrequency);
```
cw2: How about ChibiOS' macro `ST2MS()` ?
http://chibios.sourceforge.net/docs3/rt/group__time.html#ga38545158d20aed90293e73a970d66b94

josesimoes: Looks good to me!
Nice finding.

josesimoes: Actually it won't work... There are calls that pass an argument which makes this not suitable. If they were all from sysTick it would, like this it won't.

cw2: I don't understand - ST2MS macro has an argument, `return ST2MS(ticks);` (? )

josesimoes: You're right. I checked it on a forum question and it looked like there was no argument... So it fits the purpose.

PR #149, opened 2017-03-06
Improve CMake so that is not required to list each suported board name

### LINKED ISSUE #148: Update CMake so that the board name doesn't have to be listed
With the current CMake for ChibiOS  (..\CMake\Modules\FindCHIBIOS.cmake) the board name is checked to exist in the ChibiOS source and at the nf Orverlay. That is OK.
The next step is to find out the board name in a list of the supported boards. This is to determine what is the series of that board so the proper CMake files are included.
This is necessary but it requires that for each board that is added/supported the name is added to this list. Not convenient for third party boards and others that are not in the repo because it requires that this CMake file is changed to support them. Doesn't work for automated build systems, for example.

Proposed fix: remove this list from the CMake and add a new parameter to the initial CMake that specifies the series of the board. This should be a general and broad series pattern to allow it's use it any vendor/series not only STM and the current STM32 series.

josesimoes:
- now only the series name is required
- also improve CMake to support OEM boards that include the board definitions right on the target directory

Fixes #148

### DIFF
--- a/CMake/Modules/FindCHIBIOS.cmake
+++ b/CMake/Modules/FindCHIBIOS.cmake
@@ -15,79 +15,57 @@ else()
         # board found
         message(STATUS "ChibiOS board '${CHIBIOS_BOARD}' found in nanoFramework overlays")
     else()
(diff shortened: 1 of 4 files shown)


cw2:
Please resolve the conflicts.

PR #150, opened 2017-03-07
Rework linker scripts to allow definition of CLR managed heap

josesimoes:
- rework linker scripts and CMake files to make this happen
- add documentation


josesimoes:
@cw2 please take a closer look on the explanation about how to define the managed heap to make sure that the vocabulary and the concepts are technically correct.

PR #219, opened 2017-03-20
Initial work on BlockStorage per target

josesimoes:
- moved BlockStorage declaration to target platforms
- rework ChibiOS CLRStartup to target block storage
- update headers as required


josesimoes:
nice catch! 😉 
They are waiting for another attempt. The original code had a bit more stuff there. I'm trying to determine if the same structure needs to be here or not (regarding multiple blocks in the same region)

Issue #115, opened 2017-02-06
Add documentation about cmake-variants.TEMPLATE.json


josesimoes:
- Add cmake-tools-cmake-variants.md to docs\cmake
- Explain the purpose of the file, the various entries and how a developer should tweak it.
- Add link to it on the items list at the upper level main document.

MikroBusNet:
Is the following draft what you expect ?
[cmake-tools-cmake-variants.zip](https://github.com/nanoframework/nf-interpreter/files/906688/cmake-tools-cmake-variants.zip)

josesimoes:
> Is the following draft what you expect ?

yes, that is quite all right! I would just add (at the end, right after the example configuration) a list explaining what each entry there is meant for.

(in case you are not sure on any of those, just leave it empty and it will be filled in latter)

nfbot:
Kamino cloned this issue to nanoframework/Home

PR #231, opened 2017-03-24
Modified templates for easy setup

techcap:
modified launch.TEMPLATE.json for easy configuration.
add cmake-variants.TEMPLATE.ST_STM32F4_DISCOVERY.json

Signed-off-by: Bongho Lee <techcap@live.com>

### DIFF
--- a/.vscode/launch.TEMPLATE.json
+++ b/.vscode/launch.TEMPLATE.json
@@ -1,62 +1,64 @@
 {
-    "version": "0.2.0",
-    "configurations": [
-        {
-            "name": "name the launch configuration 1 (booter) here",
-            "type": "cppdbg",
-            "request": "launch",
-            "miDebuggerPath": "<path-to-gdb-insie-the-gcc-toolchain-mind-the-forward-slash>",
-            "targetArchitecture": "ARM",
-            "program": "<path-to-nanoBooter-ELF-file-in-the-build-folder-mind-the-forward-slash>",
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "nanoBooter",
+      "type": "cppdbg",
+      "request": "launch",
+      "miDebuggerPath": "C:/Program Files (x86)/GNU Tools ARM Embedded/6 2017-q1-update/bin/arm-none-eabi-gdb.exe",
+      "targetArchitecture": "ARM",
+      "program": "${workspaceRoot}/build/nanoBooter.elf",
 
-            "setupCommands": [
-                {"text": "target extended-remote localhost:3333"},
-                {"text": "monitor reset halt"},
-                {"text": "monitor flash write_image erase \"<path-to-nanoBooter-HEX-file-in-the-build-folder-mind-the-forward-slash>\" "},
-                {"text": "file <path-to-nanoBooter-ELF-file-in-the-build-folder-mind-the-forward-slash>"},
-                {"text": "monitor reset halt"}
-            ],
+      "setupCommands": [
+        { "text": "target extended-remote localhost:3333" },
+        { "text": "monitor reset halt" },
+        { "text": "monitor flash write_image erase \"${workspaceRoot}/build/nanoBooter.hex\" " },
+        { "text": "file ${workspaceRoot}/build/nanoBooter.elf" },
+        { "text": "monitor reset halt" }
+      ],
 
-            "customLaunchSetupCommands": [
-                {"text": "monitor reset halt"}
-            ],
+      "customLaunchSetupCommands": [
+        { "text": "monitor reset halt" }
+      ],
 
-            "launchCompleteCommand": "exec-continue",
-            "debugServerPath": "C:/Program Files (x86)/openocd-0.10.0/bin/openocd.exe",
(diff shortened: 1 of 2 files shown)


techcap:
Leave a comment

.vscode/launch.TEMPLATE.json
```diff
@@ -1,62 +1,64 @@
...
-            "debugServerArgs": "-s \"C:/Program Files (x86)/openocd-0.10.0/scripts/\" -f interface/<interface-config-file>.cfg -f board/<board-config-file>.cfg",
-            "serverStarted": "Info\\ :\\ [\\w\\d\\.]*:\\ hardware",
-            "filterStderr": true,
-            "externalConsole": true,
-            "cwd": "${cwd}"
-        },
+      "launchCompleteCommand": "exec-continue",
+      "debugServerPath": "C:/Program Files (x86)/openocd-0.10.0/bin/openocd.exe",
+      "debugServerArgs": "-s \"C:/Program Files (x86)/openocd-0.10.0/scripts/\" -f interface/<interface-config-file>.cfg -f board/<board-config-file>.cfg",
+      //"debugServerArgs": "-s \"C:/Program Files (x86)/openocd-0.10.0/scripts/\" -f interface/stlink-v2-1.cfg -f board/stm32f4discovery.cfg",
```
josesimoes: `//` doesn't work on json.... _comments_ aren't supported in json files

techcap: It works in VSCode, and .vscode/launch.json is used only for VSCode.

josesimoes:
Thanks for taking the time to write these, but I don't think they belong in the repo...

The templates don't have hard coded paths for a reason: each setup is different. 
Instead, they provide a place holder for a developer to replace with it's own when he/she is configuring it's environment with VS Code.
The changes you proposed for launch.TEMPLATE.json are against this.

As for providing a template cmake-variants.TEMPLATE.json specific for a reference board: I can see the value of that. The thin is that, again, there are hard coded paths there...
It's probably a better option to offer gists for all the reference boards and launch templates. This can help developers customize their owns. 
So I suggest that you create gists with these and provide links to those in a the appropriate documentation.

There is already an issue for this: #115 
Regarding the launch, there is already a document for this /docs/vscode-debug-instructions.md
Suggest that a new section is added at the bottom something "Availbale Templates " with an explanation and a list of links to gists for the various reference boards.

BTW, there are already a few of these here:

https://gist.github.com/josesimoes/c6c7e834c45ca0a112e52b9876b37814
https://gist.github.com/josesimoes/71bbd0bd90d4a47739556a4745fb3f8a
https://gist.github.com/josesimoes/c6c7e834c45ca0a112e52b9876b37814
https://gist.github.com/josesimoes/6f5916c781b390de888fd7cbd94aee5c

techcap:
I think if you stick to current status, most users will follow my steps, and they need your help. Or they will fail. Without you, I couldn't even build it. In the point of newby, I think it has to provide how to build and run through minimal work.

techcap:
I spent two days only for building this project. I think it's not the only reason that I'm newbie.

I think if you may stick to current status, most users will follow my steps, and they need your help. Or they may fail. Without you, I couldn't even build it. In the point of newbie, I think it has to provide minimum step for building and running. clone, open vs, minimal configuration, build, and run.

I don't know gist and after reading this, I started searching what is this. For newbie, separating repository will be also barrier. I think it needs some kind of quick guide.

In the case of template, current template is useless without manual. I think sample is better and easy to modify it.

Anyway, I'll respect your rule. If you don't agree, I'll close this PR.

josesimoes:
> I spent two days only for building this project. I think it's not the only reason that I'm newbie.
I think if you may stick to current status, most users will follow my steps, and they need your help. Or they may fail. Without you, I couldn't even build it. In the point of newbie, I think it has to provide minimum step for building and running. clone, open vs, minimal configuration, build, and run.

- I understand what you are saying
- I agree 200% that we need to make life easier for new comers
- documentation sure needs improvement

_but_

- hard coding paths is not a good solution... it works for you, but it doesn't for others... if they don't read the documentation anyways they'll end up frustrated because the build will fail too, just for a different reason

- [ ] we should work on a detailed step by step guide on how to setup and run a build for nanoFramework

- [ ]  we need to improve documentation

- we are here to help, that's what the community is for! 😉

MikroBusNet:
Samples could be useful, I think. Along with more detailed docs, perhaps.
The cmake.variant file is one (and the first) show-stopper. This one has to be "perfectly" documented so that it should build without errors. I mean, only compilation errors should remain, if any. Not errors related to the dev environment.

We do not have too much reference boards, so it should be pretty easy to have one sample per board. What do you think ?

josesimoes:
Regarding the cmake-variants.json we now have: 
- [templated](nf-interpreter/cmake-variants.TEMPLATE.json ) 
- [documented](https://github.com/nanoframework/Home/blob/master/docs/building/cmake-tools-cmake-variants.md)
- [gists](https://github.com/nanoframework/Home/blob/master/docs/building/cmake-tools-cmake-variants.md#templates) available for all reference targets

So, granted, that this matter could seem a bit dauting for the new comers I believe that we are in good shape now. I'm closing this one.

PR #244, opened 2017-03-25
Fix Quail nanoBooter main code

### LINKED ISSUE #178: Update nanoBooter main() in QUAIL board
nanoBooter `main()` in QUAIL board needs to be updated to follow the current workflow which is to check for nanoCLR image before moving on to the booter.

MikroBusNet:
Reworked code to comply with latest workflow.

This fixes #178 .

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/MBN_QUAIL/nanoBooter/main.c
+++ b/targets/CMSIS-OS/ChibiOS/MBN_QUAIL/nanoBooter/main.c
@@ -7,7 +7,7 @@
 #include <hal.h>
 #include <cmsis_os.h>
 
-#include <usbcfg.h>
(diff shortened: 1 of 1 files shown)


josesimoes:
@MikroBusNet out of curiosity: Why have you disable USB?

MikroBusNet:
Because there is no user interaction in my booter's code :  execution is only lasting a few seconds before it launches automatically the CLR. So there was no real need for communication at that stage. 

Do you want me to re-add that USB code ?

josesimoes:
In the current state that (remaining in booter just a short time) will happen IF there is a valid nanoCLR image. If not, or if the booter has... booted... to perform an update it will remain there _and_ will need the USB device active. So yes, please, enable USB just like on any other board.

PR #250, opened 2017-03-27
Replace types with standard C

josesimoes:
- add struct for Monitor_Ping command

### DIFF
--- a/src/CLR/Include/WireProtocol_MonitorCommands.h
+++ b/src/CLR/Include/WireProtocol_MonitorCommands.h
@@ -57,10 +57,20 @@ typedef enum MemoryMap_Options
 // backwards compatible with .NETMF
 typedef struct Monitor_Ping_Reply
 {
-    uint32_t m_source;
-    uint32_t m_dbg_flags;
+    unsigned int m_source;
+    unsigned int m_dbg_flags;
 
 }Monitor_Ping_Reply;
+
+// structure for command Monitor Ping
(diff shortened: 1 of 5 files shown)


src/CLR/Include/WireProtocol_MonitorCommands.h
```diff
@@ -57,11 +57,21 @@ typedef enum MemoryMap_Options
 // backwards compatible with .NETMF
 typedef struct Monitor_Ping_Reply
 {
-    uint32_t m_source;
-    uint32_t m_dbg_flags;
+    unsigned int m_source;
+    unsigned int m_dbg_flags;
```
cw2: Sorry for hijacking, but I don't think you really want to do this - this structure is part of a public/shared interface, so it is important to use type of exact size here, which is `uint32_t`. `int` is platform dependent and the standard does not define its exact size (it only specifies min and max values and general size rules such as sizeof(char) <= sizeof(int) <= sizeof(long) etc.).

Even though majority of platforms for which nF is being compiled most likely have 32-bit int, making the above point somewhat moot, there could be exceptions; the use of well defined types at such places is a matter of a principle (e.g. `char` may or may not be signed, `long` may or may not be of same size as `int` etc.)

josesimoes: Your comment is most valuable! Many thanks for pointing out this. 👍 
Makes me wonder if the same applies to other types that have also been replaced...

PR #326, opened 2017-05-13
Windows.Devices.Gpio POC

MikroBusNet:
This is work in progress for evaluation only.
Goal here is to check if we are going the right way of doing things regarding file locations, methods and so on. Code is really incomplete and many methods do not contain any safety checks.

This PR also needs changes in nf-class-libraries projects.


josesimoes:
Closing. Replaced with #329

PR #330, opened 2017-05-18
Chibios version 17.x changes

networkfusion:
Ready for testing.


josesimoes:
@networkfusion the automated build of Travis is failing because it's set to use ChibiOS v16.1.7.

Please edit _.travis.yml_ file and remove this switch `-DCHIBIOS_VERSION=16.1.7` from the environment settings (there are 5 entries starting at line 20). After removing that the build will use the master branch.

As you are at it, you might want to add another end there for the new  ST_STM32F769I_DISCOVERY, so a build for this targets is launched too and it gets checked. 😉

networkfusion:
@josesimoes It will likely still fail because of asm("nop") but we shall see ;-)

networkfusion:
waiting on #331

Issue #382, opened 2017-07-03
Work on Monitor_FlashSectorMap 


josesimoes:
nanoBooter is not being able to report the Flash map because the code in the command is commented in WireProtocol_MonitorCommands.c.

Uncomment and follow the implementation of CLR_DBG_Debugger::Monitor_FlashSectorMap to make it work

networkfusion:
Uncommented Monitor_FlashSectorMap in targets/CMSIS-OS/ChibiOS/nanoBooter/WireProtocol_MonitorCommands.c however it fails with BlockStorageDevice* being undefined.

MikroBusNet:
What if you include this file : nf-interpreter\src\PAL\Include\nanoPAL_BlockStorage.h ?

networkfusion:
that just causes: 'BlockStorageDevice {aka struct BLOCKSTORAGEDEVICE}' has no member named 'GetDeviceInfo'

PR #385, opened 2017-07-03
uncommented Monitor_FlashSectorMap (#382) 

networkfusion:


josesimoes:
The monitor part is coded in plain C, so some conversion/adjustment is required.
There are equivalent counterparts of the block storage in C. Please look for those to replace the failing one(s).

PR #388, opened 2017-07-07
fix for USB connection issues with F769I

networkfusion:


piwi1263:
You're sure you want the OGT1, OTG2 settings in mcuconf.h in both booter and clr different ?

josesimoes:
If there is a good reason for this, there is nothing fundamentally wrong with it...
Booter and CLR can behave differently if required. That's why they are designed to be completely separate entities. @networkfusion will tell if that's on purpose or not.

piwi1263:
That's why I asked.

networkfusion:
Kind of because the St port is listed as FS... But i have disabled it in my current code as well as changing some of the static fields to be USB2. I will push it when at my laptop. It doesn't fix the fact that deployment still fails though...

networkfusion:
hoping I have rebased correctly... please let me know if not?!

Issue #386, opened 2017-07-04
Update cmake-variants


piwi1263:
- Update the cmake variants to include the latest added options.

networkfusion:
I think this was done...

josesimoes:
Agreed! This issues wasn't referenced in the PR comment so it wasn't automatically closed.

nfbot:
Kamino cloned this issue to nanoframework/Home

Issue #312, opened 2017-05-11
Replace CMake boolean parameters with options


josesimoes:
Obvious ones are:
USE_FPU
NF_FEATURE_DEBUGGER

nfbot:
Kamino cloned this issue to nanoframework/Home

PR #389, opened 2017-07-07
updates to documentation (#386)

networkfusion:


MikroBusNet:
Not DIRECTLY related to your PR, but don't you think that we should use the same "on/off" switches for options ?
Some are using TRUE/FALSE while some others are using ON/OFF.

josesimoes:
@MikroBusNet I understand what you are saying. This is related with #312 . There are options (in CMake) that are on/off buy most of the existing ones are using true/false because the author ( 😉 ) want knowledgeable about those...

Issue #391, opened 2017-07-16
STM32F746ZG doesn't deploy


piwi1263:

piwi1263:
The block regions are not properly setup and flash/ram regions need to be changed for the time being to allow for the deployment to work and show the deployed assemblies.

nfbot:
Kamino cloned this issue to nanoframework/Home

PR #392, opened 2017-07-17
Fix regions on F746ZG (#391)

piwi1263:
- Block regions corrected
- Memory regions corrected

Set GPIO API to true in cmake-variants, build in VS Code.
Performed a full erase and deploy of generated hex binaries.
Connected an external LED on PC10 to have a small VS17 solution make a blinky.

Signed-off-by: piwi1263 <piwi1263@gmail.com>

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/ST_NUCLEO144_F746ZG/common/Device_BlockStorage.c
+++ b/targets/CMSIS-OS/ChibiOS/ST_NUCLEO144_F746ZG/common/Device_BlockStorage.c
@@ -14,7 +14,7 @@ const BlockRange BlockRange1[] =
 
 const BlockRange BlockRange2[] = 
 {
-    { BlockRange_BLOCKTYPE_CODE      ,   0, 0 }             // 08010000 nanoCLR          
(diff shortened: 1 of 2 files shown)


piwi1263:
Although trivial but for completeness:

```
using System.Threading;
using Windows.Devices.Gpio;

namespace NFApp2
{
	public class Program
    {
        // For F4-Discovery -> Off board LED is on PB7 = 1 * 16 + 7 = 23 
        // For F769I-DISCO -> LED2_GREEN is on PJ5 = 9 * 16 + 5 = 149
        // For F746ZG-NUCLEO -> Off board LED is on PC10 = 2 * 16 + 10 = 42 
        private const int LED_PIN = 42;
        private const string NAME = "Chewbacca";

        // Set up the LED
        private static GpioPin LED;

        public static void Main()
        {
            var gpio = GpioController.GetDefault();
            LED = gpio.OpenPin(LED_PIN);
            LED.SetDriveMode(GpioPinDriveMode.Output);

            for (;;)
            {
                LED.Write(GpioPinValue.High);
                Thread.Sleep(100);
                LED.Write(GpioPinValue.Low);
                Thread.Sleep(100);
            }
        }
    }
}

```

PR #428, opened 2017-08-16
Work on Devices.Spi

josesimoes:
- update code to latest version of the managed class library
- move gpioPort[] to targetPAL to allow reuse
- work on NativeTransfer for byte[] transfers (code improvement, add comments)


targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.Spi/win_dev_spi_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -273,22 +265,22 @@ HRESULT Library_win_dev_spi_native_Windows_Devices_Spi_SpiDevice::GetDeviceSelec
 
         char* retVal = "";
     #if STM32_SPI_USE_SPI1
-        strcat(retVal,"SPI1");
+        strcat(retVal,"SPI1,");
```
MikroBusNet: I do not agree with this change. 
What if the config is only using SPI1 ? You will end up with "SPI1," as a result string, not "SPI1".

josesimoes: If you are splitting this string in managed code it will work OK.

Consider that you have SPI2 and SPI enable for example.
With the previous approach it would end up on ",SPI2,SPI5" now it's "SPI2,SPI5,"

Ideally we would add the "," only as required but that would mean adding a lot of code to handle just this.

MikroBusNet: I understand.
A quick and dirty fix, for both versions of code, could then be to remove the first/last char of the resulting string if it is a "," ?

josesimoes: Yes, that would work! And it just requires a couple of lines. 👍 
Care to add that to your incoming PR? 😉

MikroBusNet: Ok. I will do that. Which version shall I use ? I would not like to remove the wrong char :D

MikroBusNet: I will take yours, because the last char of the resulting string will always be a comma ;) No need to think more than "remove the last char of that string" ;)

piwi1263: strcat(retval, retval = "" ? "SPI1" : ", SPI1") does this compute ?

josesimoes: It does, but... Won't that require an extra IF clause on each string?

MikroBusNet: @piwi1263 : Systematically removing the last char will be faster and shorter, I think ;) No need to think : remove !

PR #480, opened 2017-09-01
Spi f7

MikroBusNet:
## Description
Added specific code for STM32F7 MCU

## Motivation and Context
I am very motivated and the context is good as well ;)

## How Has This Been Tested? (if applicable)
Result not 100% accurate on Nucleo-F746ZG so far.

Note : too bad that the comments erased my initial comment :(


josesimoes:
> Note : too bad that the comments erased my initial comment :(

Know that your original comments are not erased, they remain at the top. The hint text is added afterwards. 😉

MikroBusNet:
Ok. Perhaps I did not notice. I will see with my next PR ;)

Anyway, Travis is not happy once again. Does it have the SWO option ?

josesimoes:
Travis doesn't have SWO enabled because that's a feature typically used during debug. Plus it increases RAM and FLASH usage.

It seems that you've left some SWO call in SPI. That shouldn't be there...

PR #481, opened 2017-09-01
Improve SWO config

josesimoes:
## Description
- on some ChibiOS boards the defaul for PB3 is an alternate funtion other than the RESET state which prevents the use of the pin for SWO

## How Has This Been Tested? (if applicable)
- confirmed output on SWO viewr in ST-Link with STM32F769I-DISCO board


MikroBusNet:
The other alternate can be SPI1 SCK ont F4. Not sure it should be forced to Output for all boards.

PR #493, opened 2017-09-10
Fix garbage on F7 MCUs for I2C

MikroBusNet:
Invalidate DCache before doing any transfer.


MikroBusNet:
PR #495 is doing the job.

PR #508, opened 2017-09-15
Fix compilation error for F091 with SPI

MikroBusNet:
## Description
This board does not have a _STM32_SPII2S_MAX_ #define and its frequency is limited to 12MHz.

This 12MHz frequency becomes the default value for boards other than STM32F4xx and STM32F7xx for now.

## How Has This Been Tested?
No error at compile time with SPI option ON for the STM32F091 board.

## Types of changes

## Checklist:

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.Spi/win_dev_spi_native_Windows_Devices_Spi_SpiDevice.cpp
+++ b/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.Spi/win_dev_spi_native_Windows_Devices_Spi_SpiDevice.cpp
@@ -47,10 +47,12 @@ uint16_t Library_win_dev_spi_native_Windows_Devices_Spi_SpiDevice::ComputePresca
 {
     uint16_t pre = 0;
 #if defined(STM32F4xx_MCUCONF) || defined(STM32F7xx_MCUCONF)
-    int32_t clock = STM32_SPII2S_MAX >> 1;		// SP1, SPI4, SPI5 and SPI6 on APB2
+	int32_t clock = STM32_SPII2S_MAX >> 1;		// SP1, SPI4, SPI5 and SPI6 on APB2
     if (bus == 2 || bus == 3) clock >>= 1;		// SPI2 and SPI3 on APB1
-#else
+#elif defined(STM32F0xx_MCUCONF)
     int32_t clock = 12000000 >> 1;
+#else
+    #error "Please check max SPI frequency"
 #endif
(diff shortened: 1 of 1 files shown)


targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.Spi/win_dev_spi_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -46,9 +46,12 @@ typedef Library_win_dev_spi_native_Windows_Devices_Spi_SpiConnectionSettings Spi
 uint16_t Library_win_dev_spi_native_Windows_Devices_Spi_SpiDevice::ComputePrescaler (uint8_t bus, int32_t requestedFrequency)
 {
     uint16_t pre = 0;
-	int32_t clock = STM32_SPII2S_MAX >> 1;		// SP1, SPI4, SPI5 and SPI6 on APB2
-	if (bus == 2 || bus == 3) clock >>= 1;		// SPI2 and SPI3 on APB1
-
+#if defined(STM32F4xx_MCUCONF) || defined(STM32F7xx_MCUCONF)
+    int32_t clock = STM32_SPII2S_MAX >> 1;		// SP1, SPI4, SPI5 and SPI6 on APB2
+    if (bus == 2 || bus == 3) clock >>= 1;		// SPI2 and SPI3 on APB1
+#else
```
josesimoes: Guess this should be good for all STM32F0, but might not be for other series. 
This should be stressed in the code by add an `if defined(STM32F0xx_MCUCONF)` instead of the `else`.

MikroBusNet: Sure but what about "other" MCUs we do not handle or know yet ?
Right now, I am sure only about F4/F7 and F0. That's why I have put default value to an arbitrary one.
Another solution could be to leave the code "as is" and wait a compilation error on the _STM32_SPII2S_MAX_ . In this case we will know for sure which new board we have to handle.

josesimoes: Agreed and totally understandable.
That's why I suggested that change. In order to stress that this appears to be handled "by series" and that we've tested it on F4, 7 and 0.
Adding an elseif for F0 and an else with an #error saying that this needs checking on other F's would help when that time comes.

PR #520, opened 2017-09-23
First pass on Build RTM generation

MikroBusNet:
## Description
Added a new option in the cmake-variants.json file and changed the way this option is checked inside the source files.

## How Has This Been Tested?
Compilation successfull and image size decreased accordingly.

## Types of changes

## Checklist:


MikroBusNet:
I don't understand your comment. There are no such addition of includes in this PR.
This was the goal by using the cmake-variants.json option.
Could you clarify, please ?

josesimoes:
> There are no such addition of includes in this PR. 

To make this work you have to add an `#include <target_platform.h>` to Degub.cpp.
With the current approach (and because setting the define BUILD_RTM is made in a header file) it means that that header has to be included in all code files that potentially use the RTM flag.
What I'm suggesting is, instead, adding the define to the compiler call freeing the need to tweak/include header files.

The fact that you have this added to cmake-variants takes it half way, it still needs processing and make it available in a header file as a  `#define` to be part of the build.

PR #548, opened 2017-12-17
Update targetHAL_Time.cpp

MikroBusNet:
## Description
Fixes incorrect handling of internal system time.

## Motivation and Context
- Fixes nanoframework/Home#257

## How Has This Been Tested?
With a time set internally to 12/17/2017 08:00:00, 
`Console.WriteLine("utcNow : " + DateTime.UtcNow.ToString());`
is returning 
`utcNow : 12/17/2017 08:00:39` 
The 39 seconds are due to the manual startup time of the debug session after the board's reset.

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/nanoCLR/targetHAL_Time.cpp
+++ b/targets/CMSIS-OS/ChibiOS/nanoCLR/targetHAL_Time.cpp
@@ -36,7 +36,7 @@ signed __int64  HAL_Time_CurrentDateTime(bool datePartOnly)
 
     st.wDay = (unsigned short) _dateTime.day;
     st.wMonth = (unsigned short) _dateTime.month;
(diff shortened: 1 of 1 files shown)


targets/CMSIS-OS/ChibiOS/nanoCLR/targetHAL_Time.cpp
```diff
@@ -36,7 +36,7 @@ signed __int64  HAL_Time_CurrentDateTime(bool datePartOnly)
 
     st.wDay = (unsigned short) _dateTime.day;
     st.wMonth = (unsigned short) _dateTime.month;
-    st.wYear = (unsigned short) _dateTime.year;
+    st.wYear = (unsigned short) (_dateTime.year + 1980);    // ChibiOS is counting years since 1980
```
josesimoes: @MikroBusNet nice catch!! 👏

PR #665, opened 2018-04-11
Add ADC helper class for STM32F769I

josesimoes:
## Description
- Add ADC helper class for STM32F769I
- Update AppVeyor yaml to skip build on changes to .cs files (don't make sense in this repo as they are only helper classes)

## Motivation and Context
- Improves usability by providing C# developers working with the platform with ready to use details on the ADC configuration

## Types of changes

## Checklist:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,7 @@
     files:
     - '**\AssemblyInfo.*'
     - '**/*.md'
+    - '**/*.cs'
     - 'LICENSE'
     - dir/*
     - '.gitignore'
@@ -310,6 +311,7 @@
     files:
     - '**\AssemblyInfo.*'
     - '**/*.md'
+    - '**/*.cs'
     - 'LICENSE'
     - dir/*
     - '.gitignore'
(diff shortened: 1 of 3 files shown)


josesimoes:
Focus on STM32F769I_DISCOVERY.Adc.cs 
This is my suggestion to document and provide C# developers an insight on ADC configuration for _this_ board.

The idea would be to provide such a file for every reference board. 

Plus we should have one of these for each peripheral that is not self-documenting (Serial for example, you just need to query the providers and then call the COM that you want, no need for any further details).

steffalk:
I welcome such enumerations highly, and really would see them for a) all boards supported (even if I have only a Netduino 3 so far) and b) for all things where a (newbie) developer/hobbyist would search forever for the needed information. The same is what I went through and only could succeed with the help from Peter and José about mapping a Netduino 3 I/O pin label to the actual integer number to use in C# code.

Please do include them even for cases where a µC professional wouldn't need it. Think about the entry-level hobbyists with want to _start_ there nanoframework journey and make that start as easy as possible and as not frustrating as possible.

josesimoes:
@piwi1263 and @AdrianSoundy are we good with this? can it be replicated for other targets/peripherals?

PR #669, opened 2018-04-21
Improve Esp32 debug session start

AdrianSoundy:
## Description
This change adds a delay to allow the ESP32 to reboot before the debugger starts to request details about current assemblies.  This is specific to ESP32 as using serial link.

There may be a better way to do this by monitoring the state of the CLR. Didn't seem to be possible with current flags. But this is a good stop gap change. 

## How Has This Been Tested?
Tested multiples times starting a debug session without error

## Types of changes

## Checklist:

Signed-off-by: adriansoundy<adriansoundy@gmail.com>

### DIFF
--- a/src/CLR/Debugger/Debugger.cpp
+++ b/src/CLR/Debugger/Debugger.cpp
@@ -815,10 +815,9 @@ bool CLR_DBG_Debugger::Monitor_Reboot( WP_Message* msg)
 
     WP_ReplyToCommand(msg, true, false, NULL, 0);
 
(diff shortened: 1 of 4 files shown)


AdrianSoundy:
This was only a temporary change to allow the Esp32 people to have a better experience. I was looking at a better solution so that the CLR can flag when the reboot is complete

I'll make this change so it works for all platforms

josesimoes:
Move this discussion to [Slack workspace](https://nanoframework.slack.com/archives/C4HERJAG5/p1524472150000546)

PR #679, opened 2018-05-07
Improvements in ChibiOS SPI native

josesimoes:
## Description
- Remove declared write and read buffers for SPI
- Rework ChibiOS SPI code to use sync or async drivers plus CLR events according to transaction total time lenght
- Clear SPI configuration file from all targets (not required anymore)

## Motivation and Context
- Provide "smart" call to driver's async or sync driver according to SPI transaction time length
- Remove potential limitation on having to set fixed SPI buffers at compile time

## Types of changes

## Checklist:


MikroBusNet:
Just a question : why are you setting/clearing the CS line manually ?
The way ChibiOS is handling this depends on the mode chosen in _halconf.h_  (SPI_SELECT_MODE). Currently, it defaults to "SPI_SELECT_MODE_PAD" which seems to make your CS line handling useless.

At the moment, _SPI_SELECT_MODE_ is not really handled, so I may be wrong on your intention here.

josesimoes:
> why are you setting/clearing the CS line manually ?

This was a left over from a previous experiment. It's not meant to be there. Nice catch.

piwi1263:
After first long runs, some boards with the new SPI from this PR do run only for a few minutes some for hours, but in the end the boards stop. Thanks for MikroBusNet for the help in this. It seems to be a memory issue as soon as GC starts kicking in. That is the first result, further not SPI related memory use needs to be done to make sure this PR can be ruled out as a cause for this behavior. 

Keep this PR on DO NOT MERGE, BLOCKED

MikroBusNet:
See issue # 320 : https://github.com/nanoframework/Home/issues/320
The hang may not be caused by the SPI driver.

josesimoes:
The hang is **NOT** caused by the SPI driver, but rather becuase of the GC issues.

If the board is left running with the debugger attached (OpenOCD, not managed) it eventually stops with an hard fault at random location. Always when calling something from the CLR, never when calling ChibiOS SPI driver.

PR #681, opened 2018-05-10
Remove wait from Monitor_Reboot

josesimoes:
## Description
- Remove wait from Monitor_Reboot

## Motivation and Context
- After the improvement in deploy workflow (VS extension v0.3.8.58) this is not required anymore
- Reverts #2b1f18f
- Fixes nanoFramework/Home#316

## Types of changes

## Checklist:

### DIFF
--- a/src/CLR/Debugger/Debugger.cpp
+++ b/src/CLR/Debugger/Debugger.cpp
@@ -815,10 +815,6 @@ bool CLR_DBG_Debugger::Monitor_Reboot( WP_Message* msg)
 
     WP_ReplyToCommand(msg, true, false, NULL, 0);
 
-    // Allow some time for CLR to Reboot
-    // FIXME find a better way to syncronise the reboot with clr 
-    PLATFORM_WAIT( 1000 );
-
     return true;
 }
 


josesimoes:
@AdrianSoundy please validate this.
After the improvement in deployment (VS extension v0.3.8.58) I believe that this _pause_ is not required anymore.

I've traced the root for the Console.Writeline() to this change. 
This pause is causing an awkward CLR reboot behaviour which is clearing the `c_fDebugger_Enabled` flag which, in turn, causes that the debugger output messages being skipped because the code wrongly _thinks_ that there is no debugger attached. 
Here: https://github.com/nanoframework/nf-interpreter/blob/e41c75ede58636eceeaab7529c5170641d494700/src/CLR/Diagnostics/Info.cpp#L201

piwi1263:
This is looking promising:
image

but the SWO output is gone, intended ??

josesimoes:
Not intended... A side effect of the string being " processed" in the call to the debugger output before reaching there...
I suspect that, as it is, you'll probably see one or the other.

josesimoes:
After the improvements in the debugger engine attachment brought with VS extension 0.3.8.61 this doesn't seem to be required anymore.

PR #707, opened 2018-05-27
Revert last batch of improvements on STM32 I2C code

josesimoes:
## Description
- Revert I2C code @ 99ad98a5102ea7e5b472e41a5292c7f63466aefa

## Motivation and Context
- Seems that the last batch of changes have introduced unwanted behaviour of the I2C module. This reverts the code to point of a very well tested implementation. 

**Mind to reference I2C Nuget 1.0.0-preview171**

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.I2c/win_dev_i2c_native.cpp
+++ b/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.I2c/win_dev_i2c_native.cpp
@@ -1,8 +1,3 @@
-//
-// Copyright (c) 2017 The nanoFramework project contributors
-// See LICENSE file in the project root for full license information.
-//
-
 #include "win_dev_i2c_native.h"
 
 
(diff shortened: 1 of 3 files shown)


MikroBusNet:
Simple little comment right now : **it works !**

First test was to read/write to an Eeprom and I got the expected values. Also, there were no unwanted _Read_ after a single _Write_ :

image

I will do some more tests, as deep as those done before, to confirm that we have the expected behaviour in all cases.

After that, there will still be some enhancements to be done, because some MCUs were not handled in that version, but those are minor changes, to me. Only some #defines at some point, I presume.

MikroBusNet:
To me, there were too much changes at the same time to easily tell where the flaw is/was.
But those can be kept in memory for future enhancements.

josesimoes:
I'll be pushing another commit reverting the related changes in the Core code.

MikroBusNet:
> I2C is expected to properly handle any buffer size without blocking execution 
Not really. At the moment, it does not care about buffer size. Which means that it may freeze the managed app if the buffer is large, during the time needed to transfer it completely.
This is what you had noticed with SPI as well, nothing has changed.

But what were those core changes ?

josesimoes:
The root cause of this was that the readSize and writeSize weren't being properly set when setting up the I2C operation. Fixed with nanoframework/nf-interpreter#709.

josesimoes:
> But what were those core changes ?

Everything related with CLR events for I2C operation.

josesimoes:
No need to revert the code because the reported issues where corrected with #709.

PR #717, opened 2018-06-03
Corrected the device selector string for the ESP32 serial ports

MatthiasJentsch:
## Description
The device selector string for the serial ports should be COM0,COM1,COM2 because the ESP32 has the UART_NUM_0, UART_NUM_1 and UART_NUM_2. It would be very confusing if some one should use COM1 for UART_NUM_0.

## Motivation and Context
Before the change the device selector string for the serial ports was COM1,COM2,COM3. That results in 1,2 or 3 for the portIndex field in the SerialDevice class. We use this as UART_NUM_x for the ESP32. And that's not correct because we have UART_NUM_0, UART_NUM_1 and UART_NUM_2. So it's the best to set the device selector string to COM0,COM1,COM2.

## Types of changes

## Checklist:

Signed-off-by: Matthias Jentsch <info@matthias-jentsch.de>


josesimoes:
The point on the managed API is to abstract the hardware beneath.
So, to the C# developer, it's arguable that he/she has to be/is aware of the hardware with all that detail.
Plus the UWP API uses the index starting on 1, not 0...

With this we would have a difference between platforms which is always to avoid at all costs...
So, having said that, I'm not very keen on making this change...

MatthiasJentsch:
@AdrianSoundy what do you think? We can also subtract 1 from the field SerialDevice.portIndex to get to our UART_NUM_x. That's the alternative.

AdrianSoundy:
I understand that its best to make the nanoFramework consistent. Here we were trying to make the UART numbering consistent with the numbering the ESP32 IDF uses. Currently it not one or the other.  It accepts COM0, COM1, COM2 but says in the device string COM1, COM2, COM3.

Probably best to go down the route of being more consistent with other nanoFramework platforms.

So @MatthiasJentsch  can you change it to use the portIndex - 1 as the uart number and change the com string back to what it was.

MatthiasJentsch:
We are using now portIndex -1 as uart number for the ESP32.

PR #748, opened 2018-06-15
Work on build with support for SP or DP floating point

josesimoes:
## Description
- Add CMake option DP_FLOATINGPOINT to enable DP support (default it SP)
- Add new Math functions: Abs, Max and Min
- Rework Math classes declaration
- Rework implementation to add support for new overloaded Math methods with float and double parameters
- Update declaration of mscorlib native assembly to match managed one
- Clean up native double header with unused defines for Renesas compiler

## Motivation and Context
- Closes nanoFramework/Home#311
- Closes nanoFramework/Home#312

## Types of changes

## Checklist:


src/PAL/Double/nanoPAL_NativeDouble.cpp
```diff
@@ -216,138 +195,313 @@ bool Double::IsPositiveInfinity( double d )
...
+    float Math::Tanh( float d )
+    {
+        float retVal = tanh(d); 
+        return retVal;
+    }
+
+    float Math::Truncate( float d, float& integralValue )
+    {
+        integralValue = 0;
+        float retVal = modf(d, (double*)&integralValue);
```
josesimoes: @AdrianSoundy I had to force a cast here to (double*) otherwise the build fails...
Does this means that ESP32 can use the single-precision library? Or is it something that could be missing in the compiler switches?

BTW, is the compiler set to use the FPU?

josesimoes: I meant _can't_ use the sp lib...

PR #808, opened 2018-07-26
Independent watchdog skeleton definitions 

sharmavishnu:
## Description

- These two files provide the method signatures and default weak implementations. Each MCU platform owner must use the provided method signatures and provide a strong implementation for the given MCU
- Add option to enable (and manage) MCU Watchdog 

## Motivation and Context

Enable standard API for including support for independent watchdog for any given MCU and OS platform

- Work on nanoFramework/Home#248

## How Has This Been Tested?

Used STM32F401RE nulceo board to test using the community target port. Concrete implementation of independent watchdog for the MCU STM32F401RE is also submitted in a separate PR against that repository. This will also act as a sample implementation

## Types of changes

## Checklist:

Signed-off-by: @sharmavishnu <vishnusmoke@gmail.com>


josesimoes:
@sharmavishnu great start on this! 👏 

Considering that:
1. This is a feature to be used at the core level (meaning not exposed to managed developers)
2. Keeping code size to the absolute minimum is **high** on the priority list
3. This ideally should be setup almost automatically from CMake options and/or compiler defines  

We can probably make some simplifications on what you have here.

On a first pass, these would be:

1. Drop the functions `Watchdog_SetMaxTimeout()` and `Watchdog_SetMaxTimeout(value)` 
 (the timeout should be a define)

2. Drop `Watchdog_Start()`, `Watchdog_IsRunning()` and `Watchdog_Stop()` this feature is an either YES or NO, being able to change it during run time is asking for trouble and makes the management more complicated.


Plus some of the code you have in the community PR should be added here, because it's common to the STM32 targets and it's much efficiente to have it in a single place.


Let me know your thoughts on the above! 😃

sharmavishnu:
Included all comments. The key changes are:
1. Simplified watchdog API by having only "Init" and "Reset" functions
2. Created generic watchdog implementation for STM32
3. Anyone wanting to enable watchdog, must do the following:
- Set HAL_USE_WDG as TRUE in halconf.h
- Set STM32_WDG_USE_IWDG as TRUE in mcuconf.h
- Define a new macro IWATCHDOG_CLK_TYPICAL and provide the typical clock value (in Hertz) in mcuconf.h  (see datasheet of the specific MCU)
- Define a new macro IWATCHDOG_TIMEOUT_MILLIS and provide the required watchdog timeout (in millis) in mcuconf.h
- You are now ready. Call Watchdog_Init() method to initialize and start the watchdog and call Watchdog_Reset() regularly to ensure MCU does not reset.

targets/CMSIS-OS/ChibiOS/nanoCLR/targetHAL_Watchdog.c
```diff
@@ -0,0 +1,65 @@
...
+    wdgStart(&WDGD1,&wdgConfig);   
+}
+
+/**
+ * Reset the watchdog. If not done within the timout period,
+ * the watchdog will trigger and MCU will reset
+ */ 
+void Watchdog_Reset()
+{    
+    syssts_t sts = chSysGetStatusAndLockX();
```
josesimoes: Was checking the ChibiOS test code and there is no call to `chSysGetStatusAndLockX` as you are using here...
Do we need these?

PR #835, opened 2018-09-02
Update BUILD_ESP32.md and add automated install of tools required for build

doingnz:
Create install-tools.ps1 which will install the ESP32 tools, library, IDF and openocd into default folders as used in AppVeyor CI tool. Add items to the Windows PATH and Environment to prepare VSCode to build nanoCLR. Update BUILD-ESP32.md to reflect changes and add information regarding JTAG setup (i.e. which pins to connect on ESP32 DevKitC to regular JTAG 20 pin header.

## Description
Create PowerShell script `Install-tools.ps1` that emulates portions of the appveyor.yml setup process by calling the existing Install-*.ps1 scripts to download the libraries and tools for ESP32. Add new script to also fetch compiled OpenOCD for ESP. Placeholders appveyor.yml items for STM tools. 

Note that install-tools.ps1 sets the default paths the tools are installed at. Can edit these paths before running the script for a custom install. 

Script will copy templates to required files for VSCode. VSCode will raise errors if the paths are wrong. 

Set the default paths to openOCD in .vscode/launch.TEMPLATE-ESP32.json. Added note in BUILD_ESP32.md to edit these if required. User only needs to set path to nanoCLR.elf if they use the install script. 

Added link to image of the ESP32 DevKitC pin out and 20 pin JTAG connector with wire colours on which pins to connect.

## Motivation and Context
These changes will make it much easier for new users to setup the required environment to build the nanoCLR for the first time, including hook up of JTAG. 

## How Has This Been Tested?
Delete repository. Delete environment variables. Checkout repository. Run install-tools.ps1. Launch VSCode and select the target and build all. Program flash. Debug, setp and then run code. Attach VS2017 to deploy example application. 

## Types of changes
- [X] Improvement (non-breaking change that improves a feature, code or algorithm)
- [X] Build environment change, no code change. 

## Checklist:
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have read the **CONTRIBUTING** document.

Signed-off-by: doingnz <doingnz@gmail.com>


install-tools.ps1
```diff
@@ -0,0 +1,93 @@
+$env:BOARD_NAME= 'ESP32_DEVKITC'
```
josesimoes: This should have esp32 on the make to make it absolutely clear that is meant specificaly for that target chip

.vscode/launch.TEMPLATE-ESP32.json
```diff
@@ -6,7 +6,7 @@
             "type": "cppdbg",
             "request": "launch",
             "MIMode": "gdb",
-            "miDebuggerPath": "<absolute-path-to-the-toolchain-folder-mind-the-forward-slashes>/bin/xtensa-esp32-elf-gdb.exe",
+            "miDebuggerPath": "C:/ESP32_Tools/1.22.0-80/xtensa-esp32-elf/bin/xtensa-esp32-elf-gdb.exe",
```
josesimoes: Not very keen to change this here to a fixed path... It looses the intended purpose of hey-replaced-me-with-whatever-fits-your-environment
Nothing against having that script bellow to update the local user copy though

install-tools.ps1
doingnz: I was just coming back to add notes for discussion. The file started out intended to be for all, and then that line would not be in the file. Needs to be setup in the environment in some way. Or we need to have the file platform specific and in a different folder.

.vscode/launch.TEMPLATE-ESP32.json
doingnz: I agree, especially if there is an automated tool too that can set these automatically. (I had not seen that previously in the other pending PR, which needs to also work for ESP32). They could potentially be set by the install script if it is used. If a custom folder is required, then that could be selected prior to running the install script, and then all is OK when done.

install-tools.ps1
MatthiasJentsch: I also think that this file is intended to be for all platforms. So you should set the board name in the environment.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
```diff
@@ -215,51 +263,102 @@ If you want to debug the nanoCLR on the ESP32 chip you can use the Olimex ARM-US
...
+
+
 ```
 
 You can now debug nanoCLR on the ESP32 by pressing F5 in Visual Studio Code.
 
 
 ## Notes on JTAG debugging on ESP32
 
+If flashing nanoCLR via a COM port (default), then be aware that you need to disconnect the JTAG to avoid it preventing the bootloader from running, and therefore beng unable to reprogram the ESP23. e.g. if you see the following patterin repeating, unplugh the USB-OCD-H, and then the programming will proceed.
```
MatthiasJentsch: Litte typo: "to reprogram the ESP32"

.vscode/launch.TEMPLATE-ESP32.json
MatthiasJentsch: I would agree to @josesimoes proposal. Don't put a fixed path in the template file.

MatthiasJentsch:
@doingnz Good work! Let's see if @AdrianSoundy has also comments.

josesimoes:
Gave it a bit more thought.
Having an install tools is an awesome improvement.
But it seems that this overstepping this a bit... It's heading way fast to build _something_... 
A few thoughts on this:

1. This is violating the separation of concerns principle
2. It derails on the original build system which is CMake. 
3. All the targets can presently be build from both Windows and Linux platforms, with power shell that goes away.
4. If we are not OK with CMake as the build system we can evaluate others. But, we should look into _true_ build systems, which PS is not.

So, having said all that, it's a tremendous help have a script to install the required tools. But it should stop there.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: thank you. Fixed a couple of errors in that text :-)

doingnz:
also if `install-tools.ps1`will live on, then we need to add better help for the script by following the approach seen at end of Set-PathVariable.ps1

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
```diff
@@ -22,46 +22,50 @@ The build is based on CMake tool to ease the development in all major platforms.
 You'll need:
 
 - [Visual Studio Code](http://code.visualstudio.com/)
-- [CMake](https://cmake.org/) (Minimum required version is 3.7)
```
josesimoes: Despite having a script to automate these steps we still need to provide details and alternatives to folks not using it, so the mention to CMake and Ninja should remain here.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
```diff
@@ -22,46 +22,50 @@ The build is based on CMake tool to ease the development in all major platforms.
...
-- A build system for CMake to generate the build files to. 
-  + If you have Visual Studio (full version) you can use the included NMake.
-- [Ninja](https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-win.zip). This is lightweight build system, designed for speed and it works on Windows and Linux machines. See [here](cmake/ninja-build.md) how to setup Ninja to build **nanoFramework**.
+- [CMake](https://cmake.org/download/) Download the latest stable version from and install it (Minimum required version is 3.7).
 - [Python 3.6.5](https://www.python.org/ftp/python/3.6.5/python-3.6.5.exe) Required for uploading the nanoCLR to the ESP32.
-- [OpenOCD](https://github.com/espressif/openocd-esp32/releases/download/v0.10.0-esp32-20180418/openocd-esp32-win32-0.10.0-esp32-20180418.zip) For on chip debugging of the nanoCLR
```
josesimoes: same for OpenOCD

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
```diff
@@ -22,46 +22,50 @@ The build is based on CMake tool to ease the development in all major platforms.
...
+  + If you have Visual Studio (full version) you can use the included NMake.
+ 
 ## **nanoFramework** GitHub repo
 
 If you intend to change the nanoCLR for ESP32 and create Pull Requests then you will need to fork the [nanoFramework/nf-interpreter](https://github.com/nanoFramework/nf-interpreter) to your own GitHub repo and clone the forked GitHub repo to your Windows system using an Git client such as the [GitHub Desktop application](https://desktop.github.com/).
 
 You should use the _develop_ branch for mainstream development or the _develop-network_ branch to work with the networking features which is currently a work in progress.
 
 
-## Set up CMake
```
josesimoes: This block about CMake should remain.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
```diff
@@ -215,51 +263,112 @@ If you want to debug the nanoCLR on the ESP32 chip you can use the Olimex ARM-US
...
-	"logging": {
-		"trace": true,
-		"traceResponse": true,
-		"engineLogging": true,
-		"programOutput": true,
-		"exceptions": true,
-		"moduleLoad": true
-	}
-},
+    "version": "0.2.0",
```
josesimoes: This is showing the complete launch.json file contents.
This used to be _just_ the config bit for the ESP32. I believe that's preferable to keep it that way, even if showing more than one ESP32 variant (as this is the case).
Probably an extra sentence explaining this is in order and maybe a link to a Gist with the full file.

We need to keep in mind that there are other targets (not EPS32) and the user can have other launch configurations on it's file for whatever reason. Therefore we should restrict from completely _hijacking_ it's launch file.

install-tools.ps1
```diff
@@ -0,0 +1,132 @@
+ [CmdletBinding(SupportsShouldProcess = $true)]
+ param (
+    [Parameter(Mandatory=$true,HelpMessage="Please enter one of the following NANOCLR_WINDOWS, ESP32_DEVKITC, STM32.",Position=1)][string]$BOARD_NAME = "ESP32_DEVKITC",
+	[Parameter(Mandatory=$true,HelpMessage="Please enter the COM port for NANOCLR [e.g. COM1 or /dev/ttyUSB0].",Position=2)][string]$COMPORT = "COM1",
```
josesimoes: if this is a general install, then a mandatory parameter only for ESP32 doesn't feel correct...

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: I was not sure how to switch to Rich View on PR version, so may be easier to see final version here:  https://github.com/doingnz/nf-interpreter/blob/build-esp32/targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md

CMake is listed in prerequisites (along with Visual Studio Code and Python). 

My thinking was the Prerequisites are a list of items to be installed BEFORE doing the remaining steps. As a person works through the steps that follow, there are several other downloads that are required and linked to, which we have not previously been listed in the prerequisites.  (VSCode plugins, etc ... ). Easy to put back or extend list. 

The "Setting up the build environment for ESP32 - Manual Download" describes downloading the ESP32 libraries, tools, IDF, OpenOCD, Ninja if not using the Power Shell script.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: OpenOCD, Ninja, IDF, ESP 32 Tools, pre-built libraries are both listed in the "Setting up the build environment for ESP32 - Manual Download" section.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: What if we partition the list of Prerequisites into two lists? 
1. minimum manual installed items. 
2. list of items downloaded with the script on Windows or via manual steps.

The reason I trimmed the list originally is to avoid the situation where a person downloads the items manually and then runs the install script. 

It should be made clear the Power Shell install script is for Windows, and other OS users should follow the manual steps. It is possible to make a shell script to emulate the power shell script for Linux users. (We also don't want extra work of maintain install scripts :-)

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: It was only a link to download CMake which is listed in the prerequisites. 

It raises an interesting question if the links in prerequisites should link to a product page or a download page. Previous version had product link in prerequisites lists and link to downloads in the later section. 

If we add it back, then I would suggest the CMake paragraph should be directly under Prerequisites, and also the generic setup of Visual Code and its extensions. 

The ToC of the page is now also out of date.

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: I agree that it can be simplified and that a Gist with the full file is a nice approach. I have not used Gist, so would need help. 

This specific document we are reviewing is specifically for the ESP, so I think it can be focused on ESP, but it would be good to add notes and comments about other systems as clearly nanpCLR is more than ESP32. I was thinking earlier that near the top of the file there should be mention and link to the equivalent page to this for the STM32 setup process. 

Regarding Windows vs. other operating systems, the examples of file paths and install locations are all Windows specific. `C:\...`. At some point when there is enough demand, I expect someone to step up and make a linux or other OS version of these set-up steps :-)

install-tools.ps1
josesimoes: Can you please fix this parameter being mandatory?

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
```diff
@@ -22,46 +22,50 @@ The build is based on CMake tool to ease the development in all major platforms.
 You'll need:
 
 - [Visual Studio Code](http://code.visualstudio.com/)
-- [CMake](https://cmake.org/) (Minimum required version is 3.7)
-- A build system for CMake to generate the build files to. 
-  + If you have Visual Studio (full version) you can use the included NMake.
-- [Ninja](https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-win.zip). This is lightweight build system, designed for speed and it works on Windows and Linux machines. See [here](cmake/ninja-build.md) how to setup Ninja to build **nanoFramework**.
+- [CMake](https://cmake.org/download/) Download the latest stable version from and install it (Minimum required version is 3.7).
```
josesimoes: This is now 3.11

.vscode/launch.TEMPLATE-ESP32.json
josesimoes: @doingnz can you please replace this fixed path with a place holder as the rest of the path hints?

josesimoes:
@AdrianSoundy and @MatthiasJentsch are you good with this?

Are we OK with these templates for launch and cmake-variants specific for ESP32 or those should be in the respective template, as a _unique_ file?

targets/FreeRTOS/ESP32_DevKitC/BUILD-ESP32.md
doingnz: ok, will update to 3.11

doingnz: the install_tools requires one of the 3 board types to be specified, which is why the first parameter is mandatory. i.e. must specify one of NANOCLR_WINDOWS, ESP32_DEVKITC or STM32.

Can remove mandatory from the second parameter. I had forgotten STM32 uses stlink to flash firmware .

doingnz:
I have tested with a clean machine, which prompted a fix-up for the install path for the `ESP32_TOOLCHAIN_PATH`. 

I have moved the STM32 into its on Power Shell as there was no value in them being in the same file. 

Note the zip of the tools has extra top-level folder. To avoid the extra sub-folder when extracted (which breaks the `.\install-esp32-tools.ps1` paths.), I have changed `install-esp32-toolchain.ps1` to back-up one folder level when extracting the ZIP (i.e. path has `\..` appended.)

__Need to check if this is OK for AppVeyor.yml given it shares the `install-esp32-toolchain.ps1` ?__ 

Other approaches using extract then mv could be used if there is strong opposition to the `\..` approach.

I would like to see better instructions on setting up the example JTAG added at some future date.

__Please review and hopefully this is done and dusted for now__

install-stm32-tools.ps1
```diff
@@ -0,0 +1,66 @@
...
+$buildPathExists = Test-Path "$PSScriptRoot\build\" -ErrorAction SilentlyContinue
+If($buildPathExists -eq $False)
+{
+	md $("$PSScriptRoot\build\" ) 
+
+	Write-Host "Create build folder ..."
+}
+
+Write-Host "BOARD_NAME=" $env:BOARD_NAME
+If([string]::IsNullOrEmpty($env:HEX2DFU_PATH) -or $force)
```
josesimoes: I believe this one should be `GNU_GCC_TOOLCHAIN_PATH` instead of `HEX2DFU_PATH`

install-stm32-tools.ps1
```diff
@@ -0,0 +1,66 @@
...
+{
+	$env:GNU_GCC_TOOLCHAIN_PATH='C:\GNU_Tools_ARM_Embedded\7-2017-q4-major'
+	Write-Host ("Set User Environment GNU_GCC_TOOLCHAIN_PATH='"+$env:GNU_GCC_TOOLCHAIN_PATH+"'")
+	[System.Environment]::SetEnvironmentVariable("GNU_GCC_TOOLCHAIN_PATH", $env:GNU_GCC_TOOLCHAIN_PATH, "User")
+}
+If([string]::IsNullOrEmpty($env:HEX2DFU_PATH) -or $force)
+{
+	$env:HEX2DFU_PATH= 'C:\mytools\hex2dfu'
```
josesimoes: Thinking about this hardcoded path... (and this could be me being over perfectionist) what about if the developer has it's own "tools" folder? What about automated build environments where there can exist a restriction to write/create stuff on the C: root?
Can we have this, instead, as an argument? Perhaps defaulting to this same path if nothing is provided... 

What do you think?

install-stm32-tools.ps1
```diff
@@ -0,0 +1,66 @@
...
+.SYNOPSIS
+    Install the default tools to build nanoFramework for STM32 and setup build environemnt.
+.DESCRIPTION
+    Power Shell Script to install the default tools to build nano Framework, including setting the machine path and other environment variables needed for STM32
+.EXAMPLE
+   .\install-stm32-tools.ps1 
+   Install the tools for STM32 to default path, define required envisonment variables and update VSCode files with paths and set COM19 in tasks.json
+.NOTES
+    Tested on Windows 10
+    Author:  NanoFramework Contributors
```
josesimoes: low caps 'n' here please 😀

install-stm32-tools.ps1
```diff
@@ -0,0 +1,66 @@
...
+    Created: September 9th, 2018
+.LINK
+    https://https://github.com/nanoframework/Home
+#>
```
josesimoes: git diff tools like an empty line at the end of the files, please add one here

install-esp32-tools.ps1
```diff
@@ -0,0 +1,135 @@
...
+.EXAMPLE
+   .\install-esp32-tools.ps1 -COMPORT COM19
+   Installs the tools for Espressive ESP32 to default path, define required envisonment variables and update VSCode files with paths and set COM19 in tasks.json
+.NOTES
+    Tested on Windows 10
+    Author:  NanoFramework Contributors
+    Created: September 9th, 2018
+.LINK
+    https://https://github.com/nanoframework/Home
+#>
```
josesimoes: git diff tools like an empty line at the end of the files, please add one here

Set-PathVariable.ps1
```diff
@@ -0,0 +1,82 @@
...
+   Set-PathVariable.ps1 -NewLocation "%bin%"
+   Adds the string "%bin%" to the path, when using it it is expanded.
+.NOTES
+    Tested on Windows 10 and Server 2016
+    Author:  Peter Hahndorf
+    Created: October 7th, 2016   
+.LINK
+    https://github.com/hahndorf/hacops
+    https://peter.hahndorf.eu/blog/AddingToPathVariable
+#>
```
josesimoes: git diff tools like an empty line at the end of the files, please add one here

PR #840, opened 2018-09-05
Fix Random.Next returns negative numbers #405

MatthiasJentsch:
## Description
We need to clear the first bit because rand->Next returns uint32 and the method needs to return int32.

## Motivation and Context
- Fixes nanoFramework/Home#405

## Types of changes

## Checklist:

Signed-off-by: Matthias Jentsch info@matthias-jentsch.de

### DIFF
--- a/src/CLR/CorLib/corlib_native_System_Random.cpp
+++ b/src/CLR/CorLib/corlib_native_System_Random.cpp
@@ -14,7 +14,7 @@ HRESULT Library_corlib_native_System_Random::Next___I4( CLR_RT_StackFrame& stack
 
     NANOCLR_CHECK_HRESULT(GetRandom( stack, rand ));
 
-    stack.SetResult_I4( rand->Next() );
+    stack.SetResult_I4( rand->Next() & 0x7FFFFFFF );
 
     NANOCLR_NOCLEANUP();
 }


src/CLR/CorLib/corlib_native_System_Random.cpp
```diff
@@ -14,7 +14,7 @@ HRESULT Library_corlib_native_System_Random::Next___I4( CLR_RT_StackFrame& stack
 
     NANOCLR_CHECK_HRESULT(GetRandom( stack, rand ));
 
-    stack.SetResult_I4( rand->Next() );
+    stack.SetResult_I4( rand->Next() & 0x7FFFFFFF );
```
josesimoes: Isn't simply casting a less "expensive" operation as it _just_ gets compiled without requiring any extra operations (ANDing with 0x7FFFFFFF)?

MatthiasJentsch: No. Casting for instance 0xFFFFFFFF which is a valid return value from rand->Next() into an int32 results in the value **-1**. So I think we need ANDing or modulo operation like **% 0x7FFFFFFF**

josesimoes: OK! so we have a winner! 👍

doingnz: Personally I am inclined to avoid "magic values" in code. Maybe in the future we could use `& UInt32.MaxValue` which is equivalent to the implemented `& 0x7FFFFFFF` as it helps the source code to be "self documenting" as to why the specific mask is being applied.

MatthiasJentsch: @doingnz I don't know if we have access to ```Int32.MaxValue```

josesimoes: I agree with @doingnz .
There is a constant in the CRT lib with that value...
Or maybe a comment explaining why that is being done is good enough.

PR #864, opened 2018-09-24
Updated the ifdefs to use the xx_MCUCONF as the rest of the build system for the I2C 

gligorov:
## Description
- Changed the ifdefs from using the  prior model of defs, to the current one using xx_MCUCONF as part of the name.

## Motivation and Context
- Fixes build for STM32 targets with I2C enabled.

## How Has This Been Tested?
- Got a clean copy, tested the build process after the change, and it passes with the I2C option on.

## Types of changes

## Checklist:


josesimoes:
I was thinking about this and here are my thoughts:
- The F429 has I2C enabled in the repo, how come that AppVeyor didn't error on this?
- The STM32 _NN_ XX is the define for the subfamily identifier, so it should be good enough to handle these differences in the code...
- The STM32 _NN_ XX_MCUCONF is defined at board level... we shouldn't need to get _this_ deep in order to get the differences in I2C code...

🤔

gligorov:
Thanks @josesimoes, that makes sense.  I will work with that and confirm, but was not expecting to have that much hardcoded info about the boards.  It would be good to have as much as possible of the configuration centralized for easier image creation of new boards.

josesimoes:
The plan is to **NOT** have such fine grained build options requiring specifics for each board or series. What your PR uncovered was a flaw in the script that was kept there from the very early days. 😉 

A PR is imminent improving that and, hopefully, making this one not being necessary.

gligorov:
That will be great, and I can test any new updated build options and report.

josesimoes:
@gligorov this should be fixed with #911. Please check and if it's not, let's reopen this.

gligorov:
All is good now; Got a fresh new download, and built the custom board image with no changes, and is good.

PR #990, opened 2018-11-18
Dispose fix for Issue 421

sharmavishnu:
## Types of changes

## Checklist:

Signed-off-by: @sharmavishnu

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.I2c/win_dev_i2c_native.cpp
+++ b/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Devices.I2c/win_dev_i2c_native.cpp
@@ -36,7 +36,6 @@ static const CLR_RT_MethodHandler method_lookup[] =
     NULL,
     NULL,
     NULL,
-    NULL,
     Library_win_dev_i2c_native_Windows_Devices_I2c_I2cDevice::NativeInit___VOID,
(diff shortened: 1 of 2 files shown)


josesimoes:
Can you please double check this? If no method was removed from the managed class there shouldn't be any difference in the number of method_lookup table...
If the difference is only with the checksum there is no need to update it as the I2C build takes care of that, whenever there is a change.

sharmavishnu:
1. OK. Need to learn what is "squash". Let me get back on this.
2. You are right about the number of entries in the method table. I have noticed that sometimes the number of entries change even when nothing changes in the methods. I am currently running with the recent change, else, when the managed code calls the native code, you get an exception

josesimoes:
> OK. Need to learn what is "squash". Let me get back on this.
Sure! Bing or Google for that and you'll find a bunch of docs and tutorials on this git command.

> You are right about the number of entries in the method table. I have noticed that sometimes the number of entries change even when nothing changes in the methods. I am currently running with the recent change, else, when the managed code calls the native code, you get an exception

The reason for that exception is probably that the calls are _misaligned_ because of that array that is responsible for mapping the C# calls to native functions. 
Suggest that you wipe your bin folder and do a rebuild on the I2C library project.

sharmavishnu:
Closing this after comments from @josesimoes . The branch is anyways refreshed with other changes

PR #1125, opened 2018-12-21
Improve ESP32 power down code

josesimoes:
## Description
- Improve ESP32 power down code.

## Motivation and Context
- Need to follow the new goto power down workflow.

## Types of changes

## Checklist:


AdrianSoundy:
I tried this and although code looks ok it didn't seem to work.  The only change i would make is to call the nanoHAL_Uninitialize() instead of esp_wifi_stop() before the esp_deep_sleep_start() as that should stop the wifi( not sure if it does at the moment, but it should do. Can sort that out later ).

The CPU_SetPowerMode(PowerLevel_type powerLevel) function doesn't get called. I added some tracing and it doesn't seem to exit the g_CLR_RT_ExecutionEngine.Execute()

josesimoes:
@AdrianSoundy I've pushed the changes as you've suggested. It's now calling `nanoHAL_Uninitialize_C()`.

As for not going into `SetPowerMode()` it can be either because the execution engine doesn't exit (strange because the code is common) or the flags on that if statement before are being changed after being set in `NativeStartDeepSleep___STATIC__nanoFrameworkHardwareEsp32EspNativeError()` which looks even more weird...

josesimoes:
@AdrianSoundy there was a bug in the call to `nanoHAL_Uninitialize_C()` please give this another try.

PR #1206, opened 2019-02-06
Migrate ChibiOS to v19.1

networkfusion:
TODO:
* update board_sources.cmake for H7 & L0 boards
* test all reference boards with samples
* raise an issue to swap to HAL_USE_TRNG and hardware crypto lib

- Addresses nanoFramework/Home#446.


networkfusion:
the reason for turning -Werror off is 
image

networkfusion:
The STM32F7 still has an issue with the -Werror flag re-enabled
image

networkfusion:
Whilst waiting for ChibiOS to fix the last issue stopping the STM32F769I, the only blockers are the following:
update board_sources.cmake for H7 & L0 boards
swap to HAL_USE_TRNG

Although the latter is not important (considering it will still work without it and the new crypto lib is possibly helpful as well). I am not sure how to handle the cmake for the H7 & L0 boards as the examples given in the ChibiOS platform make files are too thin... Anyone with experience with these boards is welcome to chip in with what should be changed in relation to the other platforms...

PR #1210, opened 2019-02-13
Esp32 build with Visual Studio

AdrianSoundy:
## Description
Changes so that the Esp32 build works correctly with Visual Studio

- Fixes incorrect variables
- Removes unused variables
- Changes paths to Esp32 IDF, Toolchain & LIbs so it works with same locations as used by esp32 install script.  Toolchain etc can then be shared between VSCode & VS
- Change to ESP32 CMakelist.txt so builds correctly for VScode & VS, Output correctted so it works for both. 
- Adds a get around for the problem with "cmd.exe" not found by adding the path to cmd.exe in the passed environment.  Also updated STM32 to include this parameter so there will be no need to run separate script to finish build.  The code added to handle VISUAL_STUDIO=TRUE parameter will also need to be removed in this case.

The Build directory for VS builds is organized by target device so you can build different targets without having to delete build directory which is different from VScode build.

## Motivation and Context
So it fixes ESP32 build with Visual Studio

## Types of changes

- [] Improvement (non-breaking change that improves a feature, code or algorithm)

## Checklist:


TerryFogg:
The ESP template I originally created was not tested, so it is good to see that someone with the correct environment for ESP has corrected it
Terry

PR #1221, opened 2019-02-27
Enable build, program and debug of nf-interpreter in VS2017

c-born:
## Description
This adds support for debugging the nf-interpreter in VS2017 IDE. (specifically ESP32, but can be extended to other boards)
Sample tasks are also provided for erasing and programming from the IDE.
Separate build configurations are also supported from within the IDE
 eg:
.../build/ESP32_1/...
.../build/ESP32_2...
It may be possible to extend this concept to vscode builds, but this has not been done at this stage.

## Motivation and Context
Programming from the VS2017 IDE was added recently, but debugging didn't work due to timing/sequence problems in the MIDebugEngine. This PR provides a work-around, until the problems can be addressed by upstream.

## How Has This Been Tested?
Only tested on ESP32 at this stage, although the concept should work for others by adding to the provided templates.
Still early days and some bugs to shake out! One specifically concerns breakpoints, it seems there is state retained on the J-Link (and other?) JTag interfaces, outside that of OpenOCD and GDB, such that after a time breakpoints can't be set, evenafter cloing/reopening OpenOCD and GDB.
Unplugging/plugging the JTAG interface from the USB seems to clear the problem.

## Types of changes

(Some build paths may change)

## Checklist:

Signed-off-by: David Varley <DavidAVarley@gmail.com>


targets/FreeRTOS_ESP32/ESP32_WROOM_32/CMakeLists.txt
```diff
@@ -229,6 +229,9 @@ set_property(TARGET ${NANOCLR_PROJECT_NAME}.elf    APPEND_STRING PROPERTY LINK_F
 set_property(TARGET ${NANOCLR_PROJECT_NAME}.elf    APPEND_STRING PROPERTY LINK_FLAGS " -Wl,-Map=${PROJECT_SOURCE_DIR}/build/${NANOCLR_PROJECT_NAME}.map")
 
 # Create bin file from elf
+if(NOT DEFINED EXECUTABLE_OUTPUT_PATH)
```
josesimoes: These lines here are breaking the build. They are not supposed to be required. Plain CMake builds and VS Code don't require this.

For VS Code build, after the changes in #1210, it's required to set "EXECUTABLE_OUTPUT_PATH" build option (see #1216).
(@AdrianSoundy appreciate if you could step in)

Not sure if this is required only for VS build, but it has to be addressed in another way because we have to have all build options working (plain CMake, VS and VS Code).

AdrianSoundy:
I can't see why build is failing. 
The EXECUTABLE_OUTPUT_PATH has the correct value as you can see from the file not found error message : -
- 'D:\a\1\s\build\targets\FreeRTOS_ESP32\ESP32_WROOM_32\nanoCLR.elf
Which is the same for a successful build.  Tried moving the code to start of Makelist.txt but that didn't make any difference.
So for now have removed code and it builds again.
Just need to sort out the template files and it will be good to merge.

c-born:
> Should be named launch.vs.TEMPLATE.json instead of launch.vs.SAMPLE.json to be consistent with other template files.

I'm probably not commenting in the correct place for a PR, my comments on the launch.vs etc files were on my commits, so may not be easily visible.
I said I used SAMPLE for these rather than TEMPLATE because they are not templates, which have stand-in parameters to be edited by hand or machine to turn them into the localized working file, but rather examples of a configuration that works in one particular instance (ie for me) and can then be used as a basis for localization of even a basis for a template file.
I originally included the actual localized files I was using for this, so you guys could easily test they worked for VS build and debug, my commit comment said they shouldn't be in the repo when finally merged. I think SAMPLE is a cleaner way (doesn't overwrite localized files).

AdrianSoundy:
As a result of the testing.

VS Code - Builds ok - Need to have EXECUTABLE_OUTPUT_PATH in cmake-variants but not useful to change as it just moves the executables, not the Cmake cache.

Visual Studio - Builds OK - You have 2 configs for Esp32 in the CMakeSettings.Json, best if we only have the 1. Also can you make sure the EXECUTABLE_OUTPUT_PATH has a value of "${workspaceRoot}\\Build/${name}" like the ESP32_2 config.

For the Tasks.
The tasks are being started in separate window which means if error it closes before you see whats happened. I started in a way that puts it into the Output window which seems to work better.
```
{
      "taskName": "Flash nanoCLR ESP32 COM6",
      "appliesTo": "*",
      "type": "default",
      "command": "${env.COMSPEC}",
      "args": [
        "start \"Flashing nanoCLR\" /wait python C:/ESP32_TOOLS/esp-idf-v3.1/components/esptool_py/esptool/esptool.py --chip esp32 --port \"COM6\" --baud 921600 --before \"default_reset\" --after \"hard_reset\" write_flash -z --flash_mode \"dio\" --flash_freq \"40m\" --flash_size detect 0x1000 C:/ESP32_TOOLS/libs-v3.1/bootloader.bin 0x10000 ${outDir}/nanoCLR.bin 0x8000 ${outDir}/partitions_4mb.bin"
      ]
    },
```

AdrianSoundy:
OK just leave the files as SAMPLE

josesimoes:
@c-born 

> Should be named launch.vs.TEMPLATE.json instead of launch.vs.SAMPLE.json to be consistent with other template files.

Semantics apart I guess the idea is perfectly conveyed using either name, so leave it as it is.


> I'm probably not commenting in the correct place for a PR, my comments on the launch.vs etc files were on my commits, so may not be easily visible.

Any comment that you want to make regarding the PR should be made in the github web.
You can comment specific source lines and make general comments (like this one).
Doing that on the commit messages kind of "hides" those or makes it more dificult for others reviewing the code to reach them 😉 .

c-born:
> Doing that on the commit messages kind of "hides" those or makes it more dificult for others reviewing the code to reach them  .

Yes, I'm starting to see that. I'll learn, this is a good experience!!

PR #1234, opened 2019-03-06
Add BuildNF.bat script

c-born:
## Description
Batch script to configure and build the nf-interpreter from the command line in various configurations

## Motivation and Context
While working on the VS Build I had problems with some configurations (STM32 variants) not building successfully, and went back to the simple command-line examples in the documentation.
Unfortunately these did not build successfully either, and I started on a simple batch script to save typing in all the command-line options each time I tried a changed configuration.
It grew, as these things tend to do!
The idea now is to provide script that will successfully build the various configurations as a starting point, before making changes. Various targets can be added, and the script added to as required to ensure a successful build is possible, similar to the CI pipelines.
Either the standard script or a user-modified one can be used to automate builds and tests in various configurations. The configuration built by the script can be modified using CMake-gui to quickly test other configurations.

Usage (Eg from nf-interpreter\build):

`..\BuildNF [opt] target`

  opt = -n  - Create configuration but don't run Ninja build stage
  opt = -d  - Delete target directory before configure (clean-start, deletes CMake cache)

target =  'help' or none - give help info
target = 'list' - list all targets
target = 'all' - Build all targets (using opt flags)

The first three characters of the target name are used to set the target type, currently ESP and STM, to be added to as required.

For the STM build the ChibiOS source is first looked for in the environment variable `%CHIBIOS_SOURCE%`,  next in a local repo in the same directory as the `nf-interpreter` root, and lastly via the `git clone` method. (Local repo helps with multiple builds, but it would be nice if we could actually reference the local repo directly rather than making a copy in `build` each time - something to look into) 

## How Has This Been Tested?
Tested for several ESP32 configurations and a couple of STM32 ones.
Currently the STM32Nucleo builds but the .text segment is too large for flash, hopefully area experts can tweak this for a good build, as well as adding other working configurations.

Documentation updates to come. If accepted this will be included in the getting started docs.

## Types of changes

## Checklist:

Signed-off-by: c-born <DavidAVarley@gmail.com>


josesimoes:
@c-born: granted, the command line examples in the documentation are a bit outdated.
But the CMake build from command line is working as expected. The CI build in Azure Pipelines is using those from several months now.

Please take a look at: build-chibios-stm32.yml and build-esp32.yml @ in the azure-pipelines-templates folder.

I fear that this batch file is repeating several steps that are being performed in the main CMake (for example checking the ChibiOS and ESP32 lib sources). Plus it would be another file to maintain when making changes to the build. I would rather invest on bringing the documentation up to speed with the current build options. 

But lets wait for other maintainers opinions... 😃

c-born:
@josesimoes I figured you had some pretty good magic running the pipeline builds! Not sure how I apply them to local builds? A search on .yml said I needed to install docker-compose, I use that for my Docker machine builds, do I need it here for local builds too?

With the build script I can have a local setup for every board and configuration I'm interested in, and build them all, either clean or incrementally with a single command, whenever there are any changes I want to test. (Which sounds a bit like the pipeline builds!) I was having a lot of problems getting a build to actually complete without error, and being able to have the script build the basics and let me then adjust the options via cmake-gui and see if a config would work helped a lot. Who would have thought you needed to turn SDCARD on and OneWire off if you wanted the STM32 to build? Terry said he spent days on that.

Anyway it served its primary purpose, which was to find out what was going wrong with the VS builds, so I know a bit more about that now, it all helps! I do think it will have other uses, of course I may have missed existing methods to achieve the same result, in which case I need to know what they are so I can make them more accessible in the documentation.

josesimoes:
@c-born I believe several clarifications are in order!

- When I pointed you to the Azure Pipelines yaml it wasn't my intention to suggest that you should run that locally. It was only to stress that the CMake calls are (apparently) working as expected considering that CMake is able to build successfully on Azure Pipelines and before that on AppVeyor. 
I don't think that trying to replicate an Azure Pipeline build agent locally is the way to go! So no requirement for Docker or anything even remotely close.

- Still on Azure Pipelines (and it was the same in AppVeyor) each target is built from scratch on a clean VM that Azure Pipeline takes care of spawning. There is no reuse of the previous process, building onto a folder inside the build folder or anything resembling to that. Plain simple: VM starts clean and goes through all the tasks one after the other. In the end the build artifacts are there. Simple.

- There is no requirement to set 1-Wire OFF or SDCard ON... those are build options that should work "out of the box" if they are not it's because something is missing. You can see which features are ON and OFF of the target/feature matrix that is on the nf-interpreter repo main README. The corresponding build options are being passed to CMake on Azure Pipelines.
Plus there are currently 5 different STM32 reference targets being built and not all have the same options (SD card and 1-Wire) so I don't understand why you say that you have to "turn SDCARD on and OneWire off if you wanted the STM32 to build"... 

- The build options are supposed to be documented [here](https://github.com/nanoframework/Home/blob/master/docs/building/cmake-tools-cmake-variants.md#content-explained). I'm sure those can be improved.

- I like the idea of being able to use CMake GUI to tweak a build. Never tried that myself. I don't know if our CMakes are missing anything to be able to work on that tool. If they are then we should look into fix that. The PS scripts on the repo are supposed to get the pre-requisites on the build machine, so I would say that plus using the CMake GUI should be all that is required to successfully build an image (as an alternative to VS Code build and VS build). This is why I'm having this doubts about the batch file build. Considering all the above it looks like _another_ way of seting up and launching a build. Having a 4th way to acomplish that shouldn't be needed...

c-born:
Updated with ability to automatically use/create short paths for target builds, to overcome build problems due to excessively long file names breaking Windows 250 character path limit.
(May look at extending this method to VS builds)

josesimoes:
@c-born considering that there was no more feedback on this, we should look into wrapping up this.

Somethign that caught my attention is the mismatch in the target names. If this is going to make it into the repo, the names should follow what exists today otherwise a mapping would be required and it would more confusing.

c-born:
> Somethign that caught my attention is the mismatch in the target names. If this is going to make it into the repo, the names should follow what exists today otherwise a mapping would be required and it would more confusing.

Target names were meant as generalized examples and can be anything the user wants, such as various test and stable configurations. I was keeping them slightly different from the ones used by VS2017 and vscode as I was testing more or less concurrently. I still have a couple of things to sort out with VS2017 and vscode builds which may feed back into this, so perhaps hold off a little.

What standardized and meaningful names/name patterns would you suggest?

josesimoes:
> What standardized and meaningful names/name patterns would you suggest?

The current target names, so we are coherent throughout the repos and docs. 
Unless a better idea comes up that justifies replacing all the scripts and docs mentioning them.
The pattern is basically composing the commonly used initials to designate a vendor/series followed by the reference or common name of the board/chip/soc.

No problem waiting for more work on this! 😃 just didn't want to let this look like it was forgotten or something.

c-born:
I think this is a great feature, I use it all the time!
So much easier to run a build after pulling down changes, compared with loading up VSCode or VS to do it. Delete old config that confuses the IDEs, configure and build with a single command!

PR #1273, opened 2019-03-25
Fix HAL_CONTINUATION::Dequeue_And_Execute

josesimoes:
## Description
- Fix HAL_CONTINUATION::Dequeue_And_Execute.

## Motivation and Context
- The IRQ lock/unlock and execution sequence wasn't implemented correctly.

## Types of changes

## Checklist:


josesimoes:
@AdrianSoundy could you please validate that this fix works with ESP32?
I'm having trouble debugging it...

AdrianSoundy:
I can see that we had an extra Global_Unlock which was not needed.
Now you have added calls to SystemState_SetNoLock & SystemState_ClearNoLock which were commented out in the Netmf 4.4. 
Not sure why it was commented out as it was still querying the state of SYSTEM_STATE_NO_CONTINUATIONS before calling the  HAL_CONTINUATION::Dequeue_And_Execute(). So it looks like SystemState_SetNoLock should be there. 
I suppose it is there to stop calling a continuation while in a continuation. Not sure if that's valid or not. Maybe that needs to happens.

Haven't tried debugging it yet.

josesimoes:
That's pretty much my reasoning too.
From my understanding that needs to be there to prevent cascading calls, as you've described.
Plus when it executes all the way there are two calls to "unlock" without any lock in the between (that was the aquire which was commented).
On ChibiOS that doesn't seem to matter.

When debugging TI SimpleLink with FreeRTOS those consecutive calls freeze the execution.

The difference from the current ESP32 FreeRTOS is that the lock/unlock is using a var. On the TI adaptation there isn't such var.

PR #1288, opened 2019-04-20
⏳ add helper class for ESP32 pins

Dweaver309:
## Types of changes

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/targets/FreeRTOS_ESP32/ESP32_WROOM_32/managed_helpers/ESP32_WROOM_32_PINS.cs
+++ b/targets/FreeRTOS_ESP32/ESP32_WROOM_32/managed_helpers/ESP32_WROOM_32_PINS.cs
@@ -1,0 +1,179 @@
+//
+// Copyright (c) 2018 The nanoFramework project contributors
+// See LICENSE file in the project root for full license information.
+//
+
+
+using System;
+
+namespace Pins
+{
+    public static class ESP32Device
+    {
+        public static class Gpio
+        {
+            public const int IO00 = 0;
+            public const int IO01 = 1;
+            public const int IO02 = 2;
+            public const int IO03 = 3;
+            public const int IO04 = 4;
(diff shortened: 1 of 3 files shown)


Dweaver309:
I wrote it to help me remember the default pins. Jose thought it would be
useful. If you feel it is not needed I would agree with you.
Change the code however you  want or don't include it .

Thanks for the feedback

josesimoes:
@AdrianSoundy I was the one suggesting @Dweaver309 to add this, following what is there in STM32.

I understand that EPS32 pins can be configured on-the-fly. The same happens (with less flexibility) on STM32. The point here is to have a mapping with more friendly names for the pins when one is coding.

Equivalent helper classes exist for STM32 and they are useful, specially for developers that are jumping in.

Considering the above, do you have any suggestions on making this more "developer friendly"? 😉

josesimoes:
@Dweaver309 we are giving this some thought internally and thinking on how this pin remapping can be improved even further.

Dweaver309:
Thanks for improvements in advance!

AdrianSoundy:
Started adding this to Hardware.Esp32

AdrianSoundy:
I have been looking at this but there is a problem.
The pin assignments are not fixed so you can't have a simple constant for the definition of say a SPI.Mosi pin which means you end up with a function for every pin as it has to get the current pin assignment. This will add a lot of mostly unused code to the assembly.

So what i have decided to just limit the change t a new function :-
 public static int GetFunctionPin(DeviceFunction function)

This will return the current pin assignment for any device pin.
I will extend it to include ADC which is currently missing, but these are fixed pins.

AdrianSoundy:
This can be closed once Esp32 hardware update #1392 is closed and update to 
lib-nanoFramework.Hardware.Esp32
@Dweaver309

PR #1314, opened 2019-05-14
Fixed NXP target cmakelist

MateuszKlatecki:
I have to fix NXP target CMakeList.txt because not compiling on my machine. I also simplify FindWindows.Storage.cmake

## How Has This Been Tested?
Build finished with exit code 0

## Types of changes

## Checklist:

### DIFF
--- a/CMake/Modules/FindWindows.Storage.cmake
+++ b/CMake/Modules/FindWindows.Storage.cmake
@@ -10,8 +10,6 @@ set(BASE_PATH_FOR_THIS_MODULE "${BASE_PATH_FOR_CLASS_LIBRARIES_MODULES}/Windows.
 # set include directories
 if(RTOS_CHIBIOS_CHECK)
     list(APPEND Windows.Storage_INCLUDE_DIRS "${PROJECT_BINARY_DIR}/ChibiOS_Source/ext/fatfs/src")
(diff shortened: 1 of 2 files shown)


CMake/Modules/FindWindows.Storage.cmake
```diff
@@ -10,8 +10,6 @@ set(BASE_PATH_FOR_THIS_MODULE "${BASE_PATH_FOR_CLASS_LIBRARIES_MODULES}/Windows.
 # set include directories
 if(RTOS_CHIBIOS_CHECK)
     list(APPEND Windows.Storage_INCLUDE_DIRS "${PROJECT_BINARY_DIR}/ChibiOS_Source/ext/fatfs/src")
-    list(APPEND Windows.Storage_INCLUDE_DIRS "${PROJECT_SOURCE_DIR}/targets/CMSIS-OS/ChibiOS/Include")
```
josesimoes: Not sure why this was changed. It's related with ChibiOS, so shouldn't matter to NXP and FreeRTOS…

The correct path the header files for FatFS and ChibiOS is the one that was before. Please revert.

MateuszKlatecki: I removed this line because it could be common for all of ports. On line 21 I added : list(APPEND Windows.Storage_INCLUDE_DIRS "${TARGET_BASE_LOCATION}/Include")

PR #1360, opened 2019-06-12
ToInt64 now throws exception when appropriate

MateuszKlatecki:
## Description
Convert function didn't throw exceptions if value in input string isn't correct number or is out of range and returns 0 in above situation. After this change exceptions are thrown.

## Motivation and Context

- Fixes problem with incorrect parse integers

## How Has This Been Tested?

I wrote a simple program that tests various possibilities. I can add it to example programs.
There is still problem during UInt64 and Int64 Parse. If value is out of range function didn't throw exception but return maximal or minimal allowed value.

## Types of changes

## Checklist:

### DIFF
--- a/src/CLR/CorLib/corlib_native_System_Convert.cpp
+++ b/src/CLR/CorLib/corlib_native_System_Convert.cpp
@@ -21,7 +21,7 @@ HRESULT Library_corlib_native_System_Convert::NativeToInt64___STATIC__I8__STRING
         // suport for conversion from any base
         char* endptr = NULL;
 
-        bool UInt64 = false;
+        bool isUInt64 = false;
 
         bool isSigned = (bool)stack.Arg1().NumericByRef().u1;
         long long minValue = stack.Arg2().NumericByRef().s8;
@@ -29,7 +29,7 @@ HRESULT Library_corlib_native_System_Convert::NativeToInt64___STATIC__I8__STRING
         
         // UInt64? => use also strtoull the result will be casted to Int64
         if (minValue == 0 && maxValue == 0) {
-            UInt64 = true;
+            isUInt64 = true;
             isSigned = false;
 
             //allow spaces before digits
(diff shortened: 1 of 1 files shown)


src/CLR/CorLib/corlib_native_System_Convert.cpp
```diff
@@ -19,36 +19,72 @@ HRESULT Library_corlib_native_System_Convert::NativeToInt64___STATIC__I8__STRING
 
       #if (SUPPORT_ANY_BASE_CONVERSION == TRUE)
         // suport for conversion from any base
+        char* endptr = NULL;
 
-        int64_t zero = 0;
+        bool UInt64 = false;
```
josesimoes: This var name can be confused with a type. Better to rename to... `isUInt64` for example.

src/CLR/CorLib/corlib_native_System_Convert.cpp
```diff
@@ -19,36 +19,72 @@ HRESULT Library_corlib_native_System_Convert::NativeToInt64___STATIC__I8__STRING
...
+
+        if (UInt64) {
+            // for UInt64 the check will be bypassed
+            stack.SetResult_I8(result);
+        } else if (!isSigned && maxValue != 0x7FFFFFFFFFFFFFFF && (uint64_t)result > (uint64_t)maxValue) {
+            NANOCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
+        } else if (result > maxValue || result < minValue) {
+            NANOCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
+        } else {
+            stack.SetResult_I8(result);
```
josesimoes: The `else` here is setting the return value like the `if (UInt64)`  above.
Is this the intended way? Can't they be grouped? 
A comment explaining the reasoning behind it is in order.

josesimoes:
@MateuszKlatecki yes please add some test code to the samples.
Also copy/paste the output in the PR comment above, please. Just to keep a record of the output.

src/CLR/CorLib/corlib_native_System_Convert.cpp
MateuszKlatecki: Changed

src/CLR/CorLib/corlib_native_System_Convert.cpp
MateuszKlatecki: You are right, this can be simplified.

PR #1389, opened 2019-06-28
Improvements in network configuration storing procedure

MateuszKlatecki:
## Description
Added check whether something changed in the network configuration and only in this case, write to Flash memory

## Motivation and Context

- Removal of unnecessary write operations to Flash memory

## Types of changes

## Checklist:


josesimoes:
Nice improvement!

Can we improve this a bit further?
If, besides these, we compare the new settings with the existing ones and they are the same, there is no need to go through the update as there is nothing to actually update, right?!

As it is, on targets with flash storage, it's burning the config sector over and over needlessly.

MateuszKlatecki:
You are right, we can go further and check whether any change has occurred and then save the configuration. I will try to code it.

PR #1391, opened 2019-07-01
Add sanity check for SPI dataBitLength

AdrianSoundy:
## Description
Add sanity check to dataBitLength.
If zero seems to dump out memory.

Fixes /nanoFramework/Home#497

## How Has This Been Tested?
Run provided test code.

## Types of changes

## Checklist:


targets/FreeRTOS_ESP32/ESP32_WROOM_32/nanoCLR/Windows.Devices.Spi/win_dev_spi_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -254,6 +254,7 @@ HRESULT Library_win_dev_spi_native_Windows_Devices_Spi_SpiDevice::NativeTransfer
 
         // get data bit length
         int databitLength = pConfig[ SpiConnectionSettings::FIELD___databitLength ].NumericByRef().s4;
+        if (databitLength <= 0) databitLength = 8;
```
josesimoes: @AdrianSoundy wouldn't be better to throw a bad argument exception instead of "fixing" the value?
I mean, as it is it works, of course, but the developer is not aware that he has made a mistake or set it to a wrong value.

That's the usual procedure throughout the code base and it's done in a number of places.

From what I could found (I've only worked with 8 and 16 bits) but the data can be from 8 to 16. I don't know if ESP32 can handle anything in the between...

So I suggest that the check is made for a valid number >= 8 and <= 16 (or just 8 and 16 to keep it simple) and throw an exception on invalid argument. What do you think?

PR #1401, opened 2019-07-10
Updates GCC toolchain to 8 2019 q3 

networkfusion:
- Removes depreciated structure-size-boundary
- Closes nanoFramework/home#443. 

local build still has an error

image


josesimoes:
Not sure if this can be simply removed. It's used to align the CLR structs and that's related with managed heap and gc calculations.

networkfusion:
Just going by https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#ARM-Options

image

I am not sure what can replace it though...

josesimoes:
Yeap... I'm fearing this since it popped in the last gcc update attempt... 😱
For now it's only a warning, we can always add an exclusion for it in the build. But this has to be addressed sonner rather than later.

networkfusion:
Interesting to note: https://www.mail-archive.com/gcc-patches@gcc.gnu.org/msg169032.html

It seems that the option has been removed, but it defaults to 8 anyway (which is what we use), so the option may be a red herring on the actual issue...

networkfusion:
Also, it seems that using -Wno-deprecated does not work (tried). https://stackoverflow.com/questions/56829875/how-to-suppress-warning-that-gcc-option-is-deprecated

josesimoes:
@networkfusion that's great that the "new" default is exactly what we need and, even further, that's the industry standard. 😉

josesimoes:
GCC toolchain 7z file now on Bintray.

josesimoes:
Build error with incompatible declarations fixed with #1408.

josesimoes:
Improvements in image size (SMT32F769I-DISCO)

* nanoBooter: was 10924 B is now 10900 B (less 24B)
* nanoCLR: was 427544 B is now 422940 B (less 4604B ~2%)

PR #1416, opened 2019-07-23
Update default serial baud rate

josesimoes:
## Description
- Update default serial baud rate on all targets to 921600.

## Motivation and Context
- Support change in nanoframework/nf-debugger#204.
- All current targets support this baud rate without any problem. This increases Wire Protocol throughput ~8x.
- Transport is actually performed with USB, serial is only used on very short distance on the board, meaning that no transmission issue will raise from this. Even if that's the case the WP layer has error checking mechanism that can be used to detect this, if needed.

## Types of changes

## Checklist:


AdrianSoundy:
Seeing this is a breaking change i thought i would hold off merge until i have tested it with esp32
Unless you have already tested then go ahead an merge.

josesimoes:
Yes, I've tested extensively.
It required several fixes on the debugger and also VS extension.
Was waiting for the last VS extension version for a final round of testing.

PR #1435, opened 2019-08-24
Updated .vscode templates for new setup

info-lvsys:
New development environment setup has a few incorrect settings. Missing the ESP32 settings.json which isn't very obvious for newcomers.


josesimoes:
@info-lvsys you have to use git "amend" and then "push" (with force option).

PR #1438, opened 2019-08-27
Add support for SoftAP interface for ESP32

AdrianSoundy:
## Description
This PR adds the interfaces to manage the Soft Access Point on the ESP32

- Added new configuration block for Soft AP
- Added extra flags to control the wireless interfaces better
-- Enable, AutoStart, HiddenSSID, SmartConfig
- Adds a new event when a Station connects or disconnects from AP

Due to configuration block change for wireless configuration may require change in VS extension

## Motivation and Context
- Closes nanoFramework/Home#516

## How Has This Been Tested?
Tested with a sample application that will be added to samples

## Types of changes

- [] Improvement (non-breaking change that improves a feature, code or algorithm)

## Checklist:

- [x] I have updated the documentation accordingly.


josesimoes:
Considering that:
- Wi-Fi AP is specific of Wi-Fi capable targets (currently ESP32 and TI CC3220) 
- The need for Wi-Fi AP is somewhat *niche* 

I'm thinking that adding support for Wi-Fi AP and respective configuration, regardless of the feature is used or not, adds to the image size. Which can become a problem as we keep adding stuff.
Also, because of this being made part of general Networking, the targets which are Ethernet connected only, are also taking with the Wi-Fi AP stuff.

Suggest that:
1. Make the Wi-Fi AP bits dependent of a build option.
2. The image for ESP32 includes this feature as default.
3. The VS extension is updated to properly manage the configuration blocks depending on the target having support for this or not.

AdrianSoundy:
The Wireless AP is really useful for the initial configuration of a device in combination of a small web server to handle the config requests. I am hoping to include this as part of the sample project.

I agree we need to keep the image small and don't have extra things that are not required by a platform and having a configuration option would be the best way. 

I thought made be later we can split the Wireless AP off into a separate managed code assembly to make that easier and reduce code for Ethernet only devices.

josesimoes:
Understood and that makes perfect sense. Let's have this merged and the work to split it queued up.

PR #1466, opened 2019-10-29
Enable support for multiple block storage devices

martin-kuhn:
## Description
The CLR debugger function Monitor_FlashSectorMap() is accessed in Visual Studio to detect the device capabilities and for deployment. At the moment, it does only include the first block storage device in the list, which is the internal flash of the MCU.

## Motivation and Context

This change is required to support the use of an external flash memory in addition to the internal one. Flash memory regions of the external flash have to be included in the FlashSectorMap, otherwise they will not be detected any cannot be accessed from Visual Studio.

## How Has This Been Tested?
This has successfully been tested on a STM32F427 based custom board with just the internal flash and also with the internal and external one.

## Types of changes

## Checklist:


martin-kuhn:
Finished 👍 
sorry I hadn't paid enough attention to this.

PR #1485, opened 2019-11-06
Separate out low level GPIO

AdrianSoundy:
## Description
I have created this PR for ESP32 so I can get some feedback on this change before doing any more work on it and implementing for other devices.

Separate out the GPIO so that interface can be reused by other devices instead of device going directly to low level calls. Similar to Netmf but not the same.  

Advantages
- This allows better sharing of devices.
- The device specific code can be moved to common code reducing new device porting effort. 
- Reduce memory usage if multiple devices using it.

This is currently needed for the ENC28j60 Ethernet and the Graphic touch interface. Other existing devices can also be ported to use it.

This is the first device to be done with SPI being next as also used by ENC28j60  & Graphic driver.

Problem areas:-
The current code uses an array to save all the working variables for each gpio pin where before it saved data in the managed code. This reserves memory for each possible pin even if not used.

Possible solutions:-
- Use a map for storing working data but that adds a large chuck of library code, ok if used else where.
- Use some other structure ?
- Pass in ptr to the working area when pin enabled.  It can then be decided by calling code where this is saved.

The reserved bool would have to be moved out of current array, made into a bit array.  

The current code uses a common bounce timer which I can see would be a problem if more than 1 gpio was using debounce.  i.e  example uses  Button and a limit switch.
 
Should we rename the CPU_GPIO_decl.h & cpu_gpop.cpp to something more nanoFramework ?

## Motivation and Context
Needed to move forward with other devices

## How Has This Been Tested?
ESP32 has had some testing but not extensive.

## Types of changes

## Checklist:


AdrianSoundy:
Thanks for your last feedback

I had a look at firmware sizes and for ESP32 it added about a 1K over develop branch version but i also added handling for de-bounce logic so some of that would account for that.

Also looked at speed when adding another layer. For ESP32 was around 22hz slower between current firmware and this one with a toggle loop.  I have optimised some calling code so maybe its the same again.  Did look at using inline method for CPU_GPIO_SetPinState() but couldn't get it to compile.

Now implemented a DBlinkedList for the storing the input states and separated the reserve into a bit array.  Also done some more tidying up.

I have implemented the STM32 version of cpu_gpio.ccp and now the files:-
- win_dev_gpio_native_Windows_Devices_Gpio_GpioPin.cpp
- win_dev_gpio_native_Windows_Devices_Gpio_GpioController.cpp
are common in ESP32 & STM32 builds but still in same place.

Hope to move these to a common place when all targets are done.

I haven't tested the STM32 target yet but it should build.

AdrianSoundy:
Tested on F746ZG and seems to be working ok with a couple of Output Gpio(led) and a couple of input buttons

AdrianSoundy:
I have done the NXP version now but this is very much untested as i had to add all the interrupt handling.  If you are happy with whats been done already then i suggest this is merged and i will do the NXP gpio on a separate PR so it can be tested.

josesimoes:
Got you. Let me ask @MateuszKlatecki input on this.
As for merging it, let me test this on a couple of STM and TI boars. Unless you've already done it...

AdrianSoundy:
I have just tested on ESP32 & STM32 F746ZG
Spoke to  @MateuszKlatecki  and they are happy to test NXP changes

PR #1492, opened 2019-11-11
Fixes with math classes

MateuszKlatecki:
- remove unnecessary #if (DP_FLOATINGPOINT == TRUE) conditions in R4 (float) functions
- rework Math.Internal class

## Motivation and Context

- After this change its possible to use double precision and single precision operations

## Types of changes

## Checklist:

Signed-off-by: Mateusz Klatecki <mateusz.klatecki@gc5.pl>

### DIFF
--- a/src/CLR/CorLib/corlib_native_System_MathInternal.cpp
+++ b/src/CLR/CorLib/corlib_native_System_MathInternal.cpp
@@ -5,99 +5,44 @@
 //
 #include "CorLib.h"
 
-#if !defined(NANOCLR_EMULATED_FLOATINGPOINT)
-    #include "nanoPAL_NativeDouble.h"
+HRESULT Library_corlib_native_System_MathInternal::Abs___STATIC__I4__I4( CLR_RT_StackFrame& stack )
+{
+    NATIVE_PROFILE_CLR_CORE();
+    NANOCLR_HEADER();
 
-    HRESULT Library_corlib_native_System_MathInternal::Abs___STATIC__I4__I4( CLR_RT_StackFrame& stack )
-    {
-      #if (DP_FLOATINGPOINT == TRUE)
-        stack.NotImplementedStub();  
-      #else
+    CLR_INT32 d = stack.Arg0().NumericByRefConst().s4;
+    CLR_INT32 res = abs( d );
 
-        NATIVE_PROFILE_CLR_CORE();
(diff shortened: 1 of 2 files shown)


src/CLR/System.Math/nf_native_system_math_System_Math.cpp
```diff
@@ -1139,9 +967,9 @@
 
         double d = stack.Arg0().NumericByRefConst().r8;
         double res = 0.0;
-        double retVal = System::Math::Truncate(d, res);
```
josesimoes: Nice catch here!
Can you please make this into a separate PR? Need to keep changes atomic... and this one it's clearly a bug fix...

src/CLR/CorLib/corlib_native_System_MathInternal.cpp
```diff
@@ -5,99 +5,44 @@
 //
 #include "CorLib.h"
 
-#if !defined(NANOCLR_EMULATED_FLOATINGPOINT)
```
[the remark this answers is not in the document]
MateuszKlatecki: @josesimoes but MathInternal class has nothing to do with floating points. There are only 3 functions operating on ints.
In addition, this class does not compile if DP_FLOATINGPOINT is enabled in the build options

src/CLR/System.Math/nf_native_system_math_System_Math.cpp
MateuszKlatecki: I can make this into separate PR but without this change file does not compile if DP_FLOATINGPOINT is enabled in the build options

josesimoes: Please do that. As we don't have any target building with DP FP this hasn't been catch....

src/CLR/CorLib/corlib_native_System_MathInternal.cpp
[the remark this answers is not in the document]
josesimoes: You are correct. There were those calls to FP math lib that aren't there anymore.

src/CLR/System.Math/nf_native_system_math_System_Math.cpp
MateuszKlatecki: I moved this change into separate PR

PR #1523, opened 2019-12-02
Fix simultaneous use of SDcard and USB MSD

martin-kuhn:
## Description
The Windows.Storage library allows file operations on SD cards and USB MSD. 

## Motivation and Context
With the present code, it is only possible to use either the SD card or the USB MSD but not both at the same time. To get this to work, small changes were necessary.

## How Has This Been Tested?
This has been tested on a custom STM32F427 based board with an SD card slot and an USB Host port. With the FileAccess sample from the GitHub repo, which has been adjusted to use two removable devices, the functionality could be verified.

## Types of changes

## Checklist:

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Storage/win_storage_native_Windows_Storage_StorageFolder.cpp
+++ b/targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Storage/win_storage_native_Windows_Storage_StorageFolder.cpp
@@ -72,7 +72,7 @@ HRESULT Library_win_storage_native_Windows_Storage_StorageFolder::GetRemovableSt
     char* stringBuffer;
     uint32_t driveCount = 0;
     char workingDrive[sizeof(DRIVE_PATH_LENGTH)];
-    uint16_t driveIterator = 0;
+    uint16_t maxDriveCount = 0;
 
     CLR_RT_HeapBlock* storageFolder;
     CLR_RT_TypeDef_Index storageFolderTypeDef;
@@ -80,10 +80,9 @@ HRESULT Library_win_storage_native_Windows_Storage_StorageFolder::GetRemovableSt
     CLR_RT_HeapBlock& top   = stack.PushValue();
 
   #if HAL_USE_SDC
-    bool sdCardEnumerated = false;
-  #endif
+    // increase max drive count
+    maxDriveCount++;
 
-  #if HAL_USE_SDC
     // is the SD card file system ready?
     if(sdCardFileSystemReady)
     {
@@ -93,8 +92,9 @@ HRESULT Library_win_storage_native_Windows_Storage_StorageFolder::GetRemovableSt
   #endif
 
   #if HAL_USBH_USE_MSD
-    bool usbMsdEnumerated = false;
-
+    // increase max drive count
+    maxDriveCount++;
+    
     // is the USB mass storage device file system ready?
     if(usbMsdFileSystemReady)
     {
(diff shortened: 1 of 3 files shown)


josesimoes:
@martin-kuhn have you tested this code with the following variations? 🤔 
- [ ] 1. just the SD Card
- [ ] 2. just the USB MSD
- [ ] 3. none

martin-kuhn:
Yes, I checked all of the four possibilities:
- Both SD card and USB MSD
- Just the SD card
- Just the USB MSD
- none of them
...which work as expected.

josesimoes:
Are you sure that you're not getting a Null on 3. Instead of am empty collection?
Because you moved the StorageFolder instantiation to inside the for loop, which will never get there if there are no drives.

martin-kuhn:
With no device inserted I get an Windows.Storage.StorageFolder array of length 0 (as you can see in the picture). If there's only one device, then the array has length 1 and for two devices it has length 2.
I had to move the instantiation into the for loop because otherwise it gets done only for the first pointer. At the end of the for loop, storageFolder (which is a pointer to the StorageFolder array) gets incremented and then we have to create an instance also for this one. Otherwise, the second drive would always be Null (as it was before).
[RemovableDevices](https://user-images.githubusercontent.com/54800166/70041228-0d0a4300-15bd-11ea-8d4a-f9b0c05d9935.PNG)

targets/CMSIS-OS/ChibiOS/nanoCLR/Windows.Storage/win_storage_native_Windows_Storage_StorageFolder.cpp
```diff
@@ -73,6 +73,7 @@ HRESULT Library_win_storage_native_Windows_Storage_StorageFolder::GetRemovableSt
     uint32_t driveCount = 0;
     char workingDrive[sizeof(DRIVE_PATH_LENGTH)];
     uint16_t driveIterator = 0;
+    bool busy = false;
```
josesimoes: Don't think this `busy` flag is needed. The `sdCardEnumerated` and `usbMsdEnumerated` should take care of this.
I've just tested your fix without it and it seems to be working as expected.

martin-kuhn: I tested it quickly without the busy flag. It is needed anyway. Without the busy flag, both if conditions of the SDcard and the USB MSD are true because they are both ready and not enumerated at the beginning. So they get executed both in the first iteration of the loop and then the workingDrive of the SD card gets overwritten by the workingDrive of the USB MSD.  This means that in both iterations the same device (the second one, which is the USB MSD) is mounted. So you'll end up with an array which has two StorageFolders but they are exactly the same.
I would have preferred to use the enumerated flags instead of an additional busy flag, but this is not possible because if you use only one drive, the other one is not defined.

josesimoes:
Just tested the changes with all possible variations in the PR submitted to your repo and all it's working.

image

image

image

image

martin-kuhn:
Thanks for your solution. I came to the same one yesterday, but I dropped it because of thsis:
This seems to work as long as HAL_USE_SDC is defined. You increase maxDriveCount twice for the SD-card but never for the USB MSD (by accident I guess?). In my opinion, this does not work if someone wants to use only the USB MSD. With maxDriveCount set to 1, the for loop could never reach a driveIterator with 1, so the if(driveIterator == SD_CARD_DRIVE_INDEX_NUMERIC) would never be executed. Right?

josesimoes:
That shouldn't be a problem. The SD_CARD_DRIVE_INDEX_NUMERIC is set to 1 or 0 according to the defines on what's being used.
Agreed that's not obvious that's why there is that comment trying to explain how these work. It's on the target storage header file. It was previously on the C file but I moved it there so it could be used in the rest of the code (as it is being now 😜)

martin-kuhn:
Ok, I see. So USB_MSD_DRIVE_INDEX_NUMERIC is set to 0 when HAL_USE_SDC is not defined. Great!
Maybe we should avoid that the code starting from line 167 does not get executed if none of the if-conditions are true. What do you think? That's the case when the filesystems are not ready yet.

martin-kuhn:
Ah, that's already the case. I didn't see that, sorry :)

josesimoes:
I don't think that check is required. If the file systems aren't ready the drive count will be 0 and that code block is never entered.

josesimoes:
Can you please squash all these so the changes being made here are clearer?

martin-kuhn:
I cleaned up the code with some small changes (instantiation of StorageFolder was there twice, maxDriveCount fixed). It's tested and it works in all of the 4 possibilities.
What did you mean by "squash"?

josesimoes:
> What did you mean by "squash"?

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges

martin-kuhn:
finished 🙂

PR #1530, opened 2019-12-17
Rework of sdcard storage for NXP MIMXRT1060 port

morali:
I've edited read/write binary and read/write text to open file in working thread. Which is a little bit faster and causes less stutter then previous implementation.
Also there was a problem when file write/read would end with error but the nanoCLR wouldn't get an event, which is also fixed.
## Motivation and Context

This bug caused big slowdown of application cycle while read or write was in progress.

## How Has This Been Tested?

I've tested above fix with nanoFramework storage sample and with our own application. Storage read/write were successfull and without errors.

## Types of changes

## Checklist:

Signed-off-by: morali <jeremi.jasinski@gmail.com>

### DIFF
--- a/targets/FreeRTOS/NXP/nanoCLR/Windows.Storage/win_storage_native_Windows_Storage_FileIO.cpp
+++ b/targets/FreeRTOS/NXP/nanoCLR/Windows.Storage/win_storage_native_Windows_Storage_FileIO.cpp
@@ -49,9 +49,8 @@ static void ReadTextWorkingThread(void *arg) {
   }
 
   // read string
-  if (f_gets((TCHAR *)fileIoOperation->Content, readLength, &file)) {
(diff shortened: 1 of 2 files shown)


morali:
> 
> 
> Question: any reason for replacing the platform_malloc and free?
> The correct way is to use the platform_NNNN calls.

Yes, we are using different implementation of malloc on our device.
I reverted changes to platform_malloc and platform_free.

morali:
I think there was some error with downloading some file. Could you rerun pipeline?

josesimoes:
@morali if you have a different implementation of malloc, that's OK. The platform_malloc (and free) are there for exactly that purpose: abstract whatever implementation there is for a plaform.

You have to define those @ targets\FreeRTOS\common\platform_heap.c (which I think it's already there).

PR #1536, opened 2020-01-08
Fix UpdateConfigurationBlock

martin-kuhn:
## Description
The ConfigurationManager_UpdateConfigurationBlock() method is supposed to apply changes to the existing config block. It saves the existing config block to the RAM, then it erases the corresponding flash sector and finally stores the changed config back to the flash. Because of a copy/paste error, the old config got restored instead of applying the changes.

## Motivation and Context
To make use of the update method, the source of the memcpy statement had to be changed to configurationBlock, which is the updated one.

## How Has This Been Tested?
This has been tested on a custom STM32F427 based MCU board with networking capabilities. The config block is used to save the network config.

## Types of changes

## Checklist:

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/common/targetHAL_ConfigurationManager.cpp
+++ b/targets/CMSIS-OS/ChibiOS/common/targetHAL_ConfigurationManager.cpp
@@ -344,7 +344,7 @@ __nfweak bool ConfigurationManager_UpdateConfigurationBlock(void* configurationB
             blockAddressInCopy = configSectorCopy + blockOffset;
             
             // replace config block with new content by replacing memory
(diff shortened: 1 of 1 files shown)


martin-kuhn:
Yes, that's how I understand that too. But the RAM copy never got updated with the new incoming block. Let's look at the code:
First, we search the address of the config block we want to update in the whole config sector. Then we want to copy our updated block to this address in the config sector.

// replace config block with new content by replacing memory
memcpy(blockAddressInCopy, configSectorCopy, blockSize);   <-- old one
memcpy(blockAddressInCopy, configurationBlock, blockSize);  <-- new one

It's the new updated block (configurationBlock) that we want to insert at this address in the RAM copy. The old one inserts the RAM copy into the RAM copy itself, which does not make sense to me.

PR #1538, opened 2020-01-21
FatFS fixes for F7 series

networkfusion:
## Description

Use our own diskio.c file for FatFS since the ChibiOS project is unwilling to add it.

TODO (as seperate PR's)
Use our own (up to date) FatFS repo
Add read cache invalidation to the chibios_contrib version.

## How Has This Been Tested?
Testing is still to be completed!

## Types of changes

## Checklist:

### DIFF
--- a/targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
+++ b/targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
@@ -1,0 +1,266 @@
+/*-----------------------------------------------------------------------*/
+/* Low level disk I/O module skeleton for FatFs     (C)ChaN, 2007        */
+/*-----------------------------------------------------------------------*/
+/* This is a stub disk I/O module that acts as front end of the existing */
+/* disk I/O modules and attach it to FatFs module with common interface. */
+/*-----------------------------------------------------------------------*/
+
+#include "hal.h"
+#include "ffconf.h"
+#include "diskio.h"
+
+#if HAL_USE_MMC_SPI && HAL_USE_SDC
+#error "cannot specify both MMC_SPI and SDC drivers"
+#endif
+
+#if !defined(FATFS_HAL_DEVICE)
+#if HAL_USE_MMC_SPI
+#define FATFS_HAL_DEVICE MMCD1
+#else
+#define FATFS_HAL_DEVICE SDCD1
+#endif
+#endif
+
+#if HAL_USE_MMC_SPI
+extern MMCDriver FATFS_HAL_DEVICE;
+#elif HAL_USE_SDC
+extern SDCDriver FATFS_HAL_DEVICE;
+#else
+#error "MMC_SPI or SDC driver must be specified"
+#endif
+
+#if HAL_USE_RTC
+extern RTCDriver RTCD1;
+#endif
+
+/*-----------------------------------------------------------------------*/
+/* Correspondence between physical drive number and physical drive.      */
+
+#define MMC     0
+#define SDC     0
+
+
+
+/*-----------------------------------------------------------------------*/
+/* Inidialize a Drive                                                    */
+
+DSTATUS disk_initialize (
+    BYTE pdrv         /* Physical drive number (0..) */
+)
+{
+  DSTATUS stat;
+
+  switch (pdrv) {
+#if HAL_USE_MMC_SPI
+  case MMC:
+    stat = 0;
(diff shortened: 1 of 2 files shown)


targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
```diff
@@ -0,0 +1,273 @@
...
+    cacheBufferFlush(buff, MMCSD_BLOCK_SIZE*count);
+    while (count > 0) {
+      if (mmcSequentialRead(&FATFS_HAL_DEVICE, buff))
+        return RES_ERROR;
+      buff += MMCSD_BLOCK_SIZE;
+      count--;
+    }
+    if (mmcStopSequentialRead(&FATFS_HAL_DEVICE))
+        return RES_ERROR;
+    cacheBufferInvalidate(buff, MMCSD_BLOCK_SIZE*count);
```
josesimoes: I believe this go inside the while above, right after the call to mmcSequentialRead. Depending on the data chunk size that's being read the cache issue may bite us...

targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
```diff
@@ -0,0 +1,273 @@
...
+#if HAL_USE_MMC_SPI
+  case MMC:
+    if (blkGetDriverState(&FATFS_HAL_DEVICE) != BLK_READY)
+        return RES_NOTRDY;
+    if (mmcIsWriteProtected(&FATFS_HAL_DEVICE))
+        return RES_WRPRT;
+    if (mmcStartSequentialWrite(&FATFS_HAL_DEVICE, sector))
+        return RES_ERROR;
+    // invalidate cache on buffer
+    cacheBufferFlush(buff, MMCSD_BLOCK_SIZE*count);
```
josesimoes: Same here but inside the while. Similar reason, if we don't flush the cache the data chunk to be written could be still have older data.

targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
networkfusion: So if that is the case it needs to lose the *count right...

targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
networkfusion: Same here...

networkfusion:
BTW, why is this a breaking change?

targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
josesimoes: That is correct. And now that you're mentioning it, on the code above, when it reaches the call to cacheBufferInvalidate(…), the buff pointer is already pointing to the next chunk so no invalidation would actually happen.

The correct place is right after the if and before the call to cacheBufferInvalidate(…)

targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
josesimoes: yeap

josesimoes:
> BTW, why is this a breaking change?

It's a breaking change more in the sense that's breaking with ChibiOS code. From a code perspective it looks lile more an enhacement.

targets/CMSIS-OS/ChibiOS/FatFS/fatfs_diskio.c
networkfusion: Not sure I understand... But I will commit what I think is correct and we can have another go if needed...

josesimoes:
Just edited the code to make a couple of adjustments and add proper comments.

networkfusion:
> Just edited the code to make a couple of adjustments and add proper comments.

Fine by me... I am guessing the read commands also need adding to the contrib repo for it to work as expected?

Once this is merged, I will add the changes so that we can seperate from the FatFS zip in chibiOS to using our own git repo...

networkfusion:
Also, I am guessing we need to adjust the FatFS source once separated to we can align buffers?

josesimoes:
> Also, I am guessing we need to adjust the FatFS source once separated to we can align buffers?

That's absolutely mandatory for cache invalidate otherwise will mess up with the adjacent memory.

PR #1539, opened 2020-01-23
Seperate FatFS from ChibiOS

networkfusion:
Continues to use the same version of FatFS as ChibiOS (until builds pass anyway)

TODO:
* Currently not working as cannot find ff.h/ff.c for some storage libs...
* Needs switching to a nF sanctioned repo...
* Possibly breaks NXP target until R0.14 of FatFS is used

Any help appreciated!


networkfusion:
The issue with the build (as far as I can tell) is to do with the repo (and original source for FatFS) containing `ffconf.h` which is not being overridden by the version in the targets directory. In the chibiOS zip, the file has been renamed to `_ffconf.h` so that it is disregarded.

Options:
* rename the file in our repo
* use execute_process to rename the file
* find a way to include the correct version instead...

Thoughts?

PR #1542, opened 2020-01-28
Remove -Werror from compiler flags

thopiekar:
## Description
When building stubs and introducing simple void's to be called, we get into the error about an unused variable. This one leads to an error since -Werror is enabled.
See: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html

## Motivation and Context
Together with the Chibios-based platforms we want to build stubs, which are painless to build.
The actual problem is that the generated stub includes the unused variable actually.
However, due to my leak of knowledge I don't know whether this variable is really useless or not.
But since this the current state of work, I rather see it more useful to disable -Werror for now.

## How Has This Been Tested?
Yes, but, however, on a unclean codebase.

## Types of changes
- [ ] Disabling compiler flags in the Chibios modules

## Checklist:
- [ ] See whether the CI passes. The only breakage, which might occur, is a typo somewhere.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com>

### DIFF
--- a/CMake/Modules/CHIBIOS_STM32F0xx_GCC_options.cmake
+++ b/CMake/Modules/CHIBIOS_STM32F0xx_GCC_options.cmake
@@ -17,7 +17,7 @@ set(CMAKE_EXE_LINKER_FLAGS " -Wl,--gc-sections -Wl,--no-wchar-size-warning -Wl,-
 function(NF_SET_COMPILER_OPTIONS TARGET)
 
     # include any extra options coming from any extra args?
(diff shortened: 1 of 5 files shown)


thopiekar:
Fair enough. It doesn't fix the actual problem.
At the moment I'm cleaning up my work here and try to find a way to easily set up a toolchain for many platforms.
Therefore, these kinds of workarounds are pop'ing up now.

I'll try to reproduce it, by recreating the stubs and open a new issue ticket here 👍

networkfusion:
Can this be closed as WontFix?

josesimoes:
@thopiekar any updates of this?

networkfusion:
No updates, so closing.

PR #1552, opened 2020-02-11
Fix build for NXP target

josesimoes:


josesimoes:
Can't fix. NXP build failing on FreeRTSO download after Azure Pipeline updated to git version 2.25.0.windows.1.

MateuszKlatecki:
In git 2.25.0.windows.1 version was bug (see https://github.com/git-for-windows/git/issues/2470). Now it is fixed. You have to use latest snapshot or downgrade git to earlier version.

josesimoes:
Got you. Git is part of the Azure VM so we have to wait until they update the image. Shouldn't take long. I guess...

PR #1564, opened 2020-02-25
Rework Azure Pipelines build

josesimoes:
## Description
- Add PowerShell script to setup everything and build targets.
- Add PowerShell to install CMake, OpenOCD.
- Move cmake-variants to vscode folder.
- Add PowerShell script to config VS Code.
- Add launch json to reference targets.
- Add tasks to setup build tools.
- Improve ESP32 tasks.
- Remove ESP32 and NXP specific templates (now integrated in master template).
- Rework Azure yamls to use PS1s instead of inline scripts.

## Motivation and Context
- Allow building an image from scratch with 100% automated install/setup of required tools.
- Use the same install scripts from Azure Pipelines.

## How Has This Been Tested?
- Azure Pipelines running as expected.
- PC with clean Windows install. Setup and building nf-interpreter from scripts.

## Types of changes

## Checklist:


thopiekar:
@josesimoes One idea I had in mind for a long time:
Why don't you just use the json files which are needed when developing in VS Code?
I see no reason, why you want different locations for all the options.

For example: Per supported board there would be (as already is for some boards) one cmake-variants.json. Powershell can parse json files, therefore, as the cmake-variants are the starting point for local developments, you can parse them and compile your cmake build command while giving the script the power to override the paths to your toolchain for example.

Practicing that way would ensure that all configs are located at the repo always and it would avoid the confusion in the docs that there might be cmake-variant files hosted at nf-bot, which are outdated.

What do you think?

josesimoes:
@thopiekar having the build configurations in a single location and up to date sounds great.

Bringing power shell to the build seems deviating a bit too much from CMake… or I haven't understood exactly what you're suggesting... 

Can we switch this conversation to Discord? I feel it would be much productive. 😉

PR #1566, opened 2020-02-26
Fix time conversion issues in PAL events

martin-kuhn:
## Description
Debug checks showed, that the chVTSet method is always called with a delay of zero, which should not be done.
The delay of zero was caused by dividing the ticks value by TIME_CONVERSION__TO_MILLISECONDS (=10000) which is not necessary. The compareValueTicks is already in CMSIS Ticks there, which is equal to milliseconds.
We also have to use the HAL_Time_CurrentSysTicks() method wherever the value is in CMSIS ticks instead of the HAL_Time_CurrentTime() method which returns a value in .NET ticks.

## How Has This Been Tested?
This has been tested on a custom STM32F427 based MCU board.

## Types of changes

## Checklist:


josesimoes:
Looks good on infinite loop testing.

Here's s screeshot from the scope output for a toggle on an infinite loop with 125ms.
image

josesimoes:
Running the timer sample shows some curious results:

1. Running **with** debugger attached it blinks the complete sequence as expected.

2. Running **without** debugger attached it blinks only 3 timer and then the board reboots.

🤔 this smells like a watchdog issue....

josesimoes:
Care to merge the latest changes from develop branch please so this can be merged?

martin-kuhn:
Yes, it's done.
Could you solve the timer problem?

PR #1573, opened 2020-03-05
Fix unlinking allocated GPIOs.

martin-kuhn:
## Description
Enabling chibiOS debug checks showed some error during unlinking of allocated GPIOs. The method UnlinkInputState() disables the LineEvent, performs the Unlink and frees the gpio_input_state pointer. Hence we can not use this pointer in the following statement but we have to use  gpioInputList.FirstNode() instead.

## How Has This Been Tested?
This has been tested on a custom STM32F427 based MCU board.

## Types of changes

## Checklist:


josesimoes:
@martin-kuhn can we move forward with the code review? 😃

martin-kuhn:
@josesimoes Sorry for the delay.
In my tests, this works fine. Since we do the unlinking of the gpios in a while loop with the condition that the next element can not be null, there will be one gpio left after the last iteration that won't be unlinked. Hence the gpioInputList.FirstNode() can never be null, right?

josesimoes:
Yes, but you can only call `Next()` on an existing element.
With this code on the "last" iteration you'll get a pointer to the "fist" one, which is the last.
When the execution reaches the condition and calls `Next()` it won't execute because there is no next there is only one. Meaning that the last element will remain there.

martin-kuhn:
Yes, this is exactly what I mean. But this would also be the case if we would switch to your suggestion, wouldn't it? As long as the condition of the while loop is pGpio->Next() != NULL, there will always be one remaining.

josesimoes:
So let’s have it changed then so there is none left. OK?

martin-kuhn:
what about this?

josesimoes:
@mbern could you please sign the CLA so this PR can be merged? Thank you!

PR #1580, opened 2020-03-06
Update all assemblies declaration

josesimoes:
- Following move to new MDP.


josesimoes:
Waiting for new class lib releases before publishing.

PR #1602, opened 2020-04-08
ST_STM32F769I_DISCOVERY: Remove LwIP includes on booter

thopiekar:
I see no reason why LwIP include dirs need to be included into the booter.
The booter is very basic while it shall only verify and run the CLR as far as I understood.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com>


thopiekar:
Needs some rework on the source lists.  Hope someone can confirm in short that networking is really not needed for the booter or not. Want to ensure there is no misunderstanding :)

josesimoes:
The reason why it was put there was to be used when we have the in-field update. Still not clear if this will be needed or not, but it was added there when networking got functional.
One can argue if this could/should be added only when working on it, of course, but it's there...

@thopiekar is you are on a _quest_ to clean up stuff, that's fine. But let's discuss the strategy on Discord and do this on a single PR with that goal, instead of small changes. OK? 😄

thopiekar:
Sure, let's do that. I'm just browsing through all the files around building targets and whenever I read something that I believe that is mess, then I better make the correction and commit it quickly before forgetting or reporting something trivial.

josesimoes:
Converting this PR to draft as it's not getting anywhere for some time.

josesimoes:
@thopiekar will you be looking into this or should it be closed and reopened in the future?

networkfusion:
This is closed as no further input. It can be reopened if necessary.

PR #1628, opened 2020-05-08
clang-format: Adding formatting instructions

thopiekar:
This file follows Microsoft's conventions for source files.
While thinking of the project, I think this is what we want.
However, it needs to be discussed still and can be tweaked.

Intention is to have a consistent code style and later a CI job, which ensures it is kept.
However, doing anything on the CI won't be part of the PR here.

What works for sure is auto-formatting via VisualStudio code incl. "clang format" extension. Logically, clang-format must be installed on your system. Either via %PATH% or by setting the path via extension settings.

This is an example of how a file will look like with the current settings:
https://github.com/nanoframework/nf-interpreter/pull/1628/commits/59de6fe380787dde750c3b9f3fdb63a32934721b
(Tested on my private computer - Ubuntu 20.04 - KDE)

Further info on clang-format at:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html#

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com> (github: thopiekar)


josesimoes:
Visual Studio has native support for CLang: https://devblogs.microsoft.com/cppblog/clangformat-support-in-visual-studio-2017-15-7-preview-1


But of course there is an extension for it. The same goes with VS Code.

josesimoes:
Ideally I would see this format check happening as a PR check and/or a git action. 
There are some that auto-correct the code and commit the changes, others just pass/fail.

PR #1637, opened 2020-05-18
Rework SPI to detach it allowing reuse by several devices

AdrianSoundy:
## Description, Motivation and Context
Changes to split out the SPI device layer to separate files to allow for shared use from multiple internal devices like Display, Spi Ethernet 

This code needs to be reviewed before before any other changes are made in case things need to be changed. 

The Windows.device.Spi is now common to all devices and moved to /src tree
The device layer has been split into the PAL & HAL files where the HAL only needs to ported to new devices. 

PAL -  nanoSpi.cpp / .h
HAL - CPU_SPI.cpp / CPU_SPI_decl.h

TI-SimpleLink - Changes done but noticed the SPI is a slave implementation and doesn't have any support for chip select. Not sure if it was working before.

Following nanoframework/lib-Windows.Devices.Spi#77.

## Other things to do
Modify the cmake config so you can enable the SPI without including the Windows.device.Spi
So that other devices can enable it. 

## How Has This Been Tested?
Some testing done with ESP32 but other platform untested, not sure if they even build

## Types of changes

## Checklist:


josesimoes:
Going through this!

CMakeSettings.json
```diff
@@ -1,218 +1,223 @@
 ﻿{
```
josesimoes, line 1: these changes don't belong here...

src/Windows.Devices.Spi/win_dev_spi_native.cpp.org
```diff
@@ -0,0 +1,79 @@
+﻿//
```
josesimoes, line 1: Guess this file is not needed.

src/Windows.Devices.Spi/win_dev_spi_native.h.org
```diff
@@ -0,0 +1,108 @@
+//
```
josesimoes, line 1: Guess this one either

targets/FreeRTOS_ESP32/ESP32_WROOM_32/common/WireProtocol_HAL_Interface.c
```diff
@@ -45,7 +45,7 @@ bool WP_Initialise(COM_HANDLE port)
     if ( WP_Port > UART_NUM_2 ) return false;
  
     uart_config_t uart_config = {
-        .baud_rate = 921600,                                 //baudrate
+        .baud_rate = 115200,                                 //baudrate
```
josesimoes, line 48: this change don't belong here...

AdrianSoundy: Fixed

src/Windows.Devices.Spi/win_dev_spi_native.h.org
AdrianSoundy: Fixed

src/Windows.Devices.Spi/win_dev_spi_native.cpp.org
AdrianSoundy: Fixed

CMakeSettings.json
AdrianSoundy: Restored back to original one.

src/Windows.Devices.Spi/nanoHAL_Spi.cpp
```diff
@@ -0,0 +1,308 @@
...
+// Saved config for each available SPI bus
+nanoSPI_BusConfig  spiconfig[NUM_SPI_BUSES];
+
+// Define weak functions for some optional PAL functions
+// so low level implementations don't require them.
+__nfweak bool  CPU_SPI_Initialize(int bus) { (void)bus; return true; }
+__nfweak void  CPU_SPI_Uninitialize(int bus) { (void)bus; }
+__nfweak SPI_OP_STATUS CPU_SPI_Op_Status(uint32_t deviceHandle) { (void)deviceHandle; return SPI_OP_STATUS::SPI_OP_COMPLETE; }
+
+__nfweak uint32_t CPU_SPI_Add_Device(const SPI_DEVICE_CONFIGURATION& spiDeviceConfig)
```
josesimoes, line 31: Don't understand the need for this call..

src/Windows.Devices.Spi/nanoHAL_Spi.cpp
```diff
@@ -0,0 +1,308 @@
...
+	mosiPin = GPIO_PIN_NONE;
+}
+
+//
+//  Find the Bus / Device ptr base on handle
+//
+//  return true = handle valid
+static bool getDevice(uint32_t handle, uint8_t& spiBus, int& deviceIndex)
+{
+	int type = handle >> 16 & 0x00ff;
```
josesimoes, line 69: No "magic" numbers please. If there is a call or a macro to "decode" let's use it.

src/Windows.Devices.Spi/win_dev_spi_native_Windows_Devices_Spi_SpiDevice.cpp
```diff
@@ -0,0 +1,256 @@
...
+            spiConfig.BusMode = SpiBusMode_master;
+            spiConfig.Spi_Bus = (stack.Arg1().NumericByRef().s4 - 1);
+            spiConfig.DeviceChipSelect = config[SpiConnectionSettings::FIELD___csLine].NumericByRef().s4;;
+            spiConfig.ChipSelectActive = true;  //TODO - is this something we would like to expose to managed code
+            spiConfig.SpiMode = (SpiModes)config[SpiConnectionSettings::FIELD___spiMode].NumericByRef().s4;
+            spiConfig.DataOrder16 = (DataBitOrder)config[SpiConnectionSettings::FIELD___bitOrder].NumericByRef().s4;
+            spiConfig.Clock_RateHz = config[SpiConnectionSettings::FIELD___clockFrequency].NumericByRef().s4;
+
+            // Returns deviceID or error if negative 
+            NANOCLR_CHECK_HRESULT(nanoSPI_OpenDevice(spiConfig));
```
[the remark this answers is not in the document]
AdrianSoundy: HRESULT is meant to be a negative for failure and positive for a result so it's not really against the pattern of NANOCLR_CHECK_HRESULT.

But i can make it clearer by adding an extra parameter.

src/Windows.Devices.Spi/nanoHAL_Spi.cpp
AdrianSoundy: These __nfweak calls were just placeholders in case not required by the Hal porting layer.
If that's what you meant.

AdrianSoundy:
Looks like i have hit the CreateProcess:- file not found with the ESP32 builds.
This is due to the 32K command line limit when calling CreateProcess in Windows.   By shortening path or dropping something like DAC it works but need to find a better solution as this problem also happens with graphic builds and going to be a problem with any further changes.

Supposedly you can put all command line parameters into a file to get around problem.

AdrianSoundy:
Seems we call the GCC wrapper which calls the linker. The response file seems to get expanded before calling linker therefore getting over the 32K limit on create process. Calling the linker directly seems to work.

Doesn't seem to be a way to tell gcc to just pass @file to linker without expanding.  Using -Wl,@file doesn't work

josesimoes:
@AdrianSoundy I haven't had the need to get _that_ deep on the GCC calls, so I can't provide any hints at this time.

Something worth noting: isn't this because of the old GCC version being used by ESP32 IDF? All the other targets building with the latest versions don't seem to suffer from this...

AdrianSoundy:
I tried it on the latest GCC build and had same problem.
I have got around issue now by putting all the nanoClr API sources into a separate library which removed ~200 objects from link command line. Done in separate PR #1684

AdrianSoundy:
Now that its all building again will review comments.
Remove need for ASYNC define by adding async handling too esp32 driver.

AdrianSoundy:
Think i am there now. Managed code is also committed.
Tested with ESP32 OK
Need to test with other platforms. Will do testing once i get the Discovery board.

josesimoes (approved):
LGTM

AdrianSoundy:
Tested with an SPI LCD display on both STM32F769 + ESP32

PR #1645, opened 2020-05-26
Rework ESP32 serial implementation

josesimoes:
## Description
- Add new function Esp32_SetMappedDevicePins.
- Add new enum with serial pins indexes.
- Replace "magic numbers" in serial gpio arrays with defines and enums.
- Add NF_PAL_UART struct to ESP32 serial device.
- Rework Store and Read implementations to properly handle timeout and non blocking execution.
- Fix get_BytesToRead to return count of elements in the input stream buffer.
- Remove COM1 from the available UARTS.

## Motivation and Context
- Fixes issues with using default UART GPIOs that are shared with serial flash and PSRAM.
- Addresses nanoFramework/Home#610.
- Addresses nanoFramework/Home#612.

## How Has This Been Tested?
- Serial sample. Latest version adjusted to work with ESP32.
- TX is working. 
- RX is failing.

## Types of changes

## Checklist:

### DIFF
--- a/targets/FreeRTOS_ESP32/ESP32_WROOM_32/Include/Esp32_DeviceMapping.h
+++ b/targets/FreeRTOS_ESP32/ESP32_WROOM_32/Include/Esp32_DeviceMapping.h
@@ -16,6 +16,17 @@
 #define  I2C2_DATA      25
 #define  I2C2_CLOCK     26
 
+// UART defines
+// number of pins required to configure an UART
+enum Esp32SerialPin
+{
+    Esp32SerialPin_Tx,
+    Esp32SerialPin_Rx,
+    Esp32SerialPin_Rts,
+    Esp32SerialPin_Cts,
+    Esp32SerialPin_Max,
+};
+
 enum Esp32_MapDeviceType
 {
     DEV_TYPE_GPIO,
@@ -32,5 +43,6 @@ int  Esp32_GetMappedDevicePins(Esp32_MapDeviceType DevType, int DevNumber, int P
 int  Esp32_GetMappedDevicePinsWithFunction(uint32_t alternateFunction);
 
 void Esp32_SetMappedDevicePins( uint8_t pin, int32_t alternateFunction );
+void Esp32_SetMappedDevicePins( Esp32_MapDeviceType devType, int devNumber, int8_t pinIndex, int ioPinNumber );
 
 #endif //_ESP32_DEVICEMAPPING_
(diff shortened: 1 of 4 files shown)


josesimoes:
@AdrianSoundy receiving simply doesn't work.
I've tested the code with STM32F769 and it's working. 
Sending text is not a problem. Output performs as expected.
Receiving simply does not work. I've put a logic analyser at RX pin and it picks up the chars being send from the serial terminal.

I'm using an FTDI serial to USB cable. Same as with STM32F769.
Tried several baud rates, GPIOs, UART1 & 2. Task pinned to core or not. No matter what nothing makes the UART event fire. Never. Almost sure I'm missing something obvious...

AdrianSoundy:
I have tried testing this and i am not getting any receive events. Output is working.
Maybe this is the same problem as with previous version.  Maybe related to IDF 3.3.1 will try and check.

AdrianSoundy:
Using a process of elimination found that if i put a delay for 10 secs at the end of  Native_Init and some logging in the event handler the RX data events will be fired until the delay ends and that data is picked up in managed code. After returning back to CLR the events stop. Doesn't seem to call anything in serial device to stop it. Still looking at it.

AdrianSoundy:
I have found the problem. It's due to calling uart_param_config() after driver has been installed. It resets the UART and stops the events. I have tested a change with calling individual uart_set functions and now seems to be working and we are now getting RX data :-)

There are still some other issues around dispose that cause exceptions when restarting. Still looking in to that. I don't think we need the extra RX ring buffer as the driver has already got its own ring buffer. We probably need TX ring buffer as driver one doesn't allow you to find out what has been sent and you can also disable the driver TX ring buffer.  

Probably easier if i do a PR against your repo once i am done.

josesimoes:
> UartTxWorkerTask seems to be started for each store when a long Running operation which is strings over 20 bytes at 9600. Maybe the TX thread should be like RX thread and have a queue. But then serial is so slow that it probably doesn't make a difference.

We really need this to make sure the CPU is relinqusihed back to the execution engine and the TX handled by the thread. Otherwise the C# app just freezes and this can't happen.

josesimoes:
> Do we have a unit test framework we can use for manual unit test ?

Nothing that I can recommend. We really should look into adding this...

josesimoes:
> I have found the problem. It's due to calling uart_param_config() after driver has been installed. It resets the UART and stops the events. I have tested a change with calling individual uart_set functions and now seems to be working and we are now getting RX data :-)

Awesome!! 👏 👏 👏 
Glad that you've found it. 😃 

 
> Looks like problem was caused by the latest IDF upgrade. Change in UART driver.
> 
> There are still some other issues around dispose that cause exceptions when restarting. Still looking in to that. I don't think we need the extra RX ring buffer as the driver has already got its own ring buffer. 

I'm affraid we do, otherwise it's hard to controll the read bytes requested by the Read operation... At least this is what I could find when reworking it. Keep in mind that the it's the Read operation on the C# stream that is in charge.

> We probably need TX ring buffer as driver one doesn't allow you to find out what has been sent and you can also disable the driver TX ring buffer.

Yes we need it. No way around it. Again that's because it's C# stream Store that's in charge.

> Probably easier if i do a PR against your repo once i am done.

I'll open a branch here so you can commit at will.

AdrianSoundy:
I updated driver and now working. This is now in separate PR #1658

PR #1716, opened 2020-10-02
Simplifies naming convention for RTOS targets.

networkfusion:
## Description
Currently all RTOS's use a different name for their target board
this is un-necessary as the RTOS is set.

## Motivation and Context
Optimisation of build system

## How Has This Been Tested?
If the code builds, then no apparent issues caused.

## Types of changes

## Checklist:


josesimoes:
As we are simplifying why not just make TARGET_BOARD? RTOS doesn't add any particular clarification to it...

networkfusion:
Mainly because of the distinction between some source files when searching and that use too similar names, but also for the fact that I am about to add another PR that also makes the RTOS source generic. e.g. 
image
We can simplify further later if required, I am just scared to take a step that will make it hard to support users...

networkfusion:
and would also argue that the targets name is RTOS dependent...

josesimoes:
> and would also argue that the targets name is RTOS dependent (given the fact that it uses a "target" folder).

A long long time ago we've agreed (after endless discussions) to use 'target' as the word for the concept "MCU+stuff around it".  Completely RTOS agnostic. That's why you see the current grouping: targets >> RTOS name >>  target name.
This means that you can have the same **target** available on several platforms/RTOSes (you can find that on one of the experimental branches with a STM32L0 nucleo board).

josesimoes:
Plus: we already have the concept of 'TARGET_SERIES' which is, correctly, non related with RTOS or anything else apart from the... SERIES of the target... 😄

networkfusion:
When the ChibiOS 20x PR is merged, this current PR will contain even more conflicts. Given that most of the changes are simple renames, it would probably be easier to start again. This (closed) PR can be used as a reference.

PR #1736, opened 2020-10-17
Update Azure pipeline to include ESP WROVER KIT with ILI9341

AdrianSoundy:
## Description
Test build for Wrover V4 target

## Types of changes

## Checklist:


AdrianSoundy:
Using same board name probably just creates same target binary
How do we use different parameters but use same esp32 directory ?

josesimoes:
We have a problem here... The board name matches the location where it goes into the Bintray repo for distribution. It will overwrite the "regular" ESP32 image. Need an alternative name. Maybe adding an _UI suffix?

josesimoes:
> Using same board name probably just creates same target binary
> How do we use different parameters but use same esp32 directory ?

That usage scenario wasn't been taken into account on the existing build system. 
Clearly it needs to be. Have to do some rework to accommodate this.

PR #1754, opened 2020-11-02
Fix Read/Write in System.Device.I2c

AdrianSoundy:
## Description

Fixes issues when a null parameter is passed into nativeTransmit() method.
As the parameter is a SpanByte structure we don't get a null parameter but a SpanByte structure with a null array

Also found an issue with the I2cTransferResult return type.  If Windows.Devices.I2c assembly not loaded then it couldn't find type.
Added new I2cTransferResult for System.Device.I2C and renamed variable for one used by Windows.Devices.I2c
m_I2cTransferResult ->  m_I2cTransferResult_old.

The Windows.Devices.I2c I2cTransferResult  can be removed once no longer used.

## Motivation and Context
- Fixes nanoFramework/Home#659

## How Has This Been Tested?

Tested on ESP32 only with nf_ssd1306 supplied test program for writes
and a BMP280 test program for reads

## Types of changes

## Checklist:

### DIFF
--- a/src/CLR/Core/TypeSystem.cpp
+++ b/src/CLR/Core/TypeSystem.cpp
@@ -2838,7 +2838,8 @@ static const TypeIndexLookup c_TypeIndexLookup[] = {
 
     TIL("System.Net.Sockets", "SocketException", m_SocketException),
 
-    TIL("Windows.Devices.I2c", "I2cTransferResult", m_I2cTransferResult),
+    TIL("System.Device.I2c", "I2cTransferResult", m_I2cTransferResult),
+    TIL("Windows.Devices.I2c", "I2cTransferResult", m_I2cTransferResult_old),
 
 #if (HAL_USE_ESP32_RMT_OPTION == TRUE)
     TIL("nanoFramework.Hardware.Esp32.Rmt", "RmtCommand", m_RmtCommand),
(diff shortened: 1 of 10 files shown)


josesimoes:
Notes for future reference.

The root cause for this "misbehaviour" is that `SpanByte` it's a struct (value type) therefore can't be assigned a `null`.
In this case when we were checking with "is the object null?" is, in reality, if it was assigned (which it wasn't obviously).

To be syntactically correct we would have to:
1) allow nullable types 
2) declare `SpanByte` as nullable 
3) check is `SpanByte.HasValue`

The fix in this PR works because the `_array` field is `null` on a "new" `SpanByte` which is what's actually being passed in the stack at Arg1 position, after the call to `ldnull` to load a _pseudo null_ `SpanByte`. 

cc @Ellerbach @krwq

Ellerbach:
> The fix in this PR works because the `_array` field is `null` on a "new" `SpanByte`

Yes, that's the expected behavior. The way to check if there is something is on `_length`, if `_length == 0` then the array is empty
So for the array is empty, then don't try to get it out as most likely (but not necessary) the `_array` is null

josesimoes (approved):
With no further comments, let's have this merged.
Thanks for the quick fix!

PR #1760, opened 2020-11-08
Corrects the Orgpal3 cmake-variants file

networkfusion:
## Description
CAN and OneWire should not be on by default.
Added the configuration files for CAN and OneWire
(although not working by default).

## Motivation and Context
Confusion with default build options.

## How Has This Been Tested?
No changes that would affect default build.
More likely to build if default options are incorrect.

## Types of changes

## Checklist:


josesimoes:
@networkfusion these two features are not enabled on ORGPAL on purpose. 
The above can't work because the pins used in the proposed configuration for CAN and 1-Wire overlap signals used in the SD Card connector and the flash chips on the board.

networkfusion:
@josesimoes , That Is why I had them set to false. I have removed the offending code, but kept the files with a warning comment instead.

Note that the main thing is that the variants has been corrected as the peripherals were on by default.

josesimoes (approved):

PR #1787, opened 2020-11-20
Makes RTOS version parameter generic

networkfusion:
## Description
Helps with new ChibiOS targets that might use different (future) release.

## Types of changes

## Checklist:
- [ ] I have read the [CONTRIBUTING](https://github.com/nanoframework/.github/blob/master/CONTRIBUTING.md) document.


josesimoes (changes requested):
You have both RTOS_VERSION_TAG and RTOS_VERSION which seems wrong.

I would say that the TAG suffix is not helpful and makes one wonder a bit why it's there.
I mean RTOS_VERSION is perfectly fine and raises no questions. Opposed to RTOS_VERSION_TAG on which one wonders "what is tag" or "what exactly is this"?

It's there just because it can be a Git *tag* but it could also be a Git *branch* or if the version control is not Git at all it could be something else.

So I'm OK with changing it to a generic RTOS agnostic variable name RTOS_VERSION but the tag suffix is definitely not belonging there.

networkfusion:
@josesimoes my changes only copy the way it was done for freertos. There must be a reason for the additional check done which requires 2 names for the same parameter and a check to make sure the one being taken from the variants file not being an empty string. The question is, why the check was needed in the first place?

josesimoes:
> @josesimoes my changes only copy the way it was done for freertos. There must be a reason for the additional check done which requires 2 names for the same parameter and a check to make sure the one being taken from the variants file not being an empty string. The question is, why the check was needed in the first place?

Then need to figure out why is that being done first. No point keep doing something wrong, if that's the case.

PR #1796, opened 2020-11-30
Enc28J60 driver only

AdrianSoundy:
## Description

ENC28J60 driver only and integrated to build

Adds the NF_NETWORKING_ENC28J60 option to include the ENC28J60 driver but only if networking included.
Uses the generic SPI and GPIO interfaces so good for all platforms.

GPIO_INTERRUPT_SERVICE_ROUTINE for gpio updated as it needed the user arg added back. Driver required it for passing the * NetIF. Updated all references in all platforms.

Currently this is the Driver only and not linked to any platforms. So can't be used yet.

If you link this at startup, the call to Time_Initialise() needs to move before the Network initialisation because the open driver uses a Continuation. For testing temporary moved into TargetHal.cpp.  This was on ESP32, didn't check on other platforms. 

Driver is not included in any of the current Azure builds.
 
Todo:-

The System.Net needs to be updated so that the enc28j60 adapter can be added/initialised from the C# app using the correct SPI parameters. Will create separate task for this.
 
## How Has This Been Tested?

The only testing done was to make sure the driver starts up properly and changes in IO working (On ESP32)
Tested build on ESP32 & STM32F769I_DISCOVERY.

## Types of changes

## Checklist:


CMakeLists.txt
```diff
@@ -500,7 +503,11 @@ if( API_System.Net OR
         set(USE_SECURITY_MBEDTLS_OPTION TRUE CACHE INTERNAL "NF security MBEDTLS")
     endif()
 
-    if(API_Windows.Devices.Wifi)
+    if(NF_NETWORKING_ENC28J60)
+        set(USE_ENC28J60_DRIVER_OPTION TRUE CACHE INTERNAL "NF ENC28j60 network driver")
```
josesimoes, line 507: The else statement should be here too setting USE_ENC28J60_DRIVER_OPTION to false, otherwise we may end up with USE_ENC28J60_DRIVER_OPTION undefined.

josesimoes (commented):
All the rest LGTM!

CMakeLists.txt
AdrianSoundy: Done, will merge

PR #1824, opened 2021-01-02
StorageFile DateTime Heap corruption fix

AdrianSoundy:
## Description
When we updated a dateTime in the StorageFile object it corrupted the heap because code assumes its its referencing a DateTime field but in this case it is a DATA_OBJECT with reference to DATATYPE_VALUETYPE of class DateTime 

Changed code to use the corlib function-
Library_corlib_native_System_DateTime::GetValuePtr( CLR_RT_HeapBlock& ref ) method to get ptr to DateTime
This now seems to work OK

So far only done for ESP32 will probably need same change for other platforms. 

Please review, if ok i'll do other platforms

## Motivation and Context
reported by @TheFLC on Discord

## How Has This Been Tested?
Simple test program crashed on Heap compact, haven't as yet run samples
```
		public static void Main()
		{
			Debug.WriteLine("Hello from nanoFramework!");
			{

				var InternalDevices = Windows.Storage.KnownFolders.InternalDevices;
				StorageFolder spiffDevice = InternalDevices.GetFolders()[0];

				StorageFile sf = spiffDevice.CreateFile("serial.txt", CreationCollisionOption.ReplaceExisting);

                                Debug.WriteLine($"dateTime:{sf.DateCreated}");

				sf.Delete();
			}

			Debug.WriteLine("CreateFile");

			nanoFramework.Runtime.Native.GC.Run(true);

			Debug.WriteLine("Heap compacted");

		}
```

## Types of changes

## Checklist:

### DIFF
--- a/azure-pipelines-templates/pack-publish-managed-helpers.yml
+++ b/azure-pipelines-templates/pack-publish-managed-helpers.yml
@@ -57,6 +57,6 @@ steps:
       command: push
       nuGetFeedType: external
       packagesToPush: '$(Build.ArtifactStagingDirectory)/*.nupkg'
-      publishFeedCredentials: 'NuGet'
+      publishFeedCredentials: 'NuGet-$(System.TeamProject)'
     condition: and( succeeded(), eq(variables['BUILD_NUGET_PACKAGE'], true), ne(variables['Build.Reason'], 'PullRequest') )
     displayName: Push managed helpers NuGet packages to NuGet
--- a/azure-pipelines-templates/pack-publish-ti-sl-managed-helpers.yml
+++ b/azure-pipelines-templates/pack-publish-ti-sl-managed-helpers.yml
@@ -57,6 +57,6 @@ steps:
       command: push
       nuGetFeedType: external
       packagesToPush: '$(Build.ArtifactStagingDirectory)/*.nupkg'
-      publishFeedCredentials: 'NuGet'
+      publishFeedCredentials: 'NuGet-$(System.TeamProject)'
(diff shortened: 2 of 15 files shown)


andybeetz:
I also found a fix by using:

dateFieldRef.SetInteger((CLR_UINT64)GetFileTimeFromPath(workingPath));

Is there much difference between the two approaches?

josesimoes (approved):
Makes perfect sense! Good job.

josesimoes:
> I also found a fix by using:
> 
> dateFieldRef.SetInteger((CLR_UINT64)GetFileTimeFromPath(workingPath));
> 
> Is there much difference between the two approaches?
> 
> Also! StorageFile.cpp also needs the fix.

That works too. But Adrian approach has a better check for and dealing with null pointers and such.
Not that that is expected to happen, but it can.

josesimoes:
> Also! StorageFile.cpp also needs the fix.
@AdrianSoundy when you're porting this to the other platforms please take this into account.

AdrianSoundy:
I tested the approach of 
`dateFieldRef.SetInteger((CLR_UINT64)GetFileTimeFromPath(workingPath));`
But it doesn't work. You don't get a heap problem but if you try to print out dataTime from storageFile object it gives you wrong type exception.
@andybeetz @josesimoes

josesimoes (approved):
LGTM!

AdrianSoundy:
NXP is using a different method to save datetime by creating an instance of DateTime updating ticks then saving it to object ref.
Probably works but i'll update it anyway so all platforms consistent. Also moving code to separate method reduces code size.

PR #1936, opened 2021-05-29
Improvements in convert and format handling

edleno2:
## Description

- Added FormatException and code to report formatting errors in Parse routines, Updated versions and interop structure.
- Replace some current exceptions with FormatException to be consistent with .net core.
- Replaced some other generic exceptions with ArgumentOutOfRangeException to be consistent with .net core.
- Some cleanup generated by CLANG format as headers were modified to allow new exception type.
- Fixed handling of signed and unsigned Parse strings - allowed "-0" for example.  Refactored some code to allow this.
- Added errno processing for integers to catch formatting errors from native math functions.
- Corrected min/max check logic for integer and float logic.
- Correctly handle (.net core compliant) empty strings and/or space preceding string values about to be parsed.
- Report cast errors as invalid cast - .net core compliant.
- Added System.IO.Ports to STM32F429I-Discovery device in variants.make so it would generate correct includes when running cmake.

## Motivation and Context

- Fixes nanoFramework/Home#708

## How Has This Been Tested?

Ran unit tests on Win32 with both SUPPORT_ANY_BASE_CONVERSION and without that, then ESP32-WROOM and STM32F429I-Discovery.  
## Types of changes

## Checklist:


edleno2:
Trying to find another way to use errno - the lwip library has overridden the functionality of errno processing and I'm not sure how to replace it AND have it work.  I can fix the build, but then the unit tests fail when run on hardware.

src/CLR/CorLib/corlib_native.cpp
```diff
@@ -1480,7 +1486,7 @@ const CLR_RT_NativeAssemblyData g_CLR_AssemblyNative_mscorlib =
 
 #if (NANOCLR_REFLECTION == TRUE)
 
-    0x5032B8FF,
+    0x1AA8C441,
 
 #elif (NANOCLR_REFLECTION == FALSE)
```
josesimoes, line 1492: Missing the update for the assembly without reflection. See comment in CorLib.

targets/ChibiOS/ST_STM32F429I_DISCOVERY/cmake-variants.json
```diff
@@ -53,6 +53,7 @@
           "API_System.Device.I2c": "ON",
           "API_Windows.Devices.Pwm": "ON",
           "API_Windows.Devices.SerialCommunication": "ON",
+          "API_System.IO.Ports": "ON",
```
josesimoes, line 56: Although correct, unrelated with this PR. Please revert.
We should keep these "atomic".

src/CLR/Core/CLR_RT_HeapBlock.cpp
```diff
@@ -893,7 +893,7 @@ HRESULT CLR_RT_HeapBlock::PerformUnboxing(const CLR_RT_TypeDef_Instance &cls)
 
     if (this->DataType() != DATATYPE_OBJECT)
     {
-        NANOCLR_SET_AND_LEAVE(CLR_E_WRONG_TYPE);
+        NANOCLR_SET_AND_LEAVE(CLR_E_INVALID_CAST);
```
josesimoes, line 896: Are we sure that it's safe to replace this exception? I'm asking because I recall this being used in the generics processing and not sure about the impact at this time... So, just asking.

josesimoes:
@edleno2 just rebased and fix some style issues. As I couldn't push to your fork, I've made this in branch "pr1936". 
Please pull from it.

josesimoes:
@edleno2 just fixed the build issue with `errno`. Pull from `pr1936` 😉

src/CLR/CorLib/corlib_native.cpp
edleno2: Updated list of source files in noReflection project and copy/pasted new checksum.

src/CLR/Core/CLR_RT_HeapBlock.cpp
edleno2: The current hResult of CLR_E_WRONG_TYPE generates a generic Exception, so any managed code doing a Catch (Exception ex) will continue to work.  I could not find any references to the CLR_E_WRONG_TYPE in nanoCLR that was looking at that specific HR and acted different - i.e. all of the references were to setting that hResult.  The invalid cast exception IS what is being returned in .net core in the same situation, so that led me to the change after checking if there were references to the old hr.

targets/ChibiOS/ST_STM32F429I_DISCOVERY/cmake-variants.json
edleno2: Reverted

src/CLR/CorLib/corlib_native.cpp
josesimoes: It had to both checksum and the method lookup array. Did you do that with the "new" one after adding the linked file?

edleno2:
Closing this PR - made a mess of it while trying to rebase.  Will open a new one.

PR #1938, opened 2021-05-31
Improvements in convert and format handling

edleno2:
## Description

- Added FormatException and code to report formatting errors in Parse routines, Updated versions and interop structure.
- Replace some current exceptions with FormatException to be consistent with .net core.
- Replaced some other generic exceptions with ArgumentOutOfRangeException to be consistent with .net core.
- Some cleanup/code-style of lists in exceptions header file.
- Fixed handling of signed and unsigned Parse strings - allowed "-0" for example. Refactored some code to allow this.
- Added errno processing for integers to catch formatting errors from native math functions.
- Corrected min/max check logic for integer and float logic.
- Correctly handle (.net core compliant) empty strings and/or space preceding string values about to be parsed.
- Report cast errors as invalid cast - .net core compliant.

## Motivation and Context

- Fixes nanoFramework/Home#708

## How Has This Been Tested?

Ran unit tests on Win32 with both SUPPORT_ANY_BASE_CONVERSION and without that, then ESP32-WROOM and STM32F429I-Discovery.
## Types of changes

## Checklist:


src/CLR/CorLib/corlib_native_System_Convert.cpp
```diff
@@ -7,6 +7,7 @@
 #include "corlib_native.h"
 #include <ctype.h>
 #include <base64.h>
+#include <cerrno> // when running with lwip the use of errno is affected by _REENT_ONLY - see below.  LWIP is included via corlib_native.h and lower (HAL).  Win32 does not use it
```
edleno2, line 10: This include is actually needed for win32, but it is here for any device that the lwip headers are not "active".  See more comments below.

src/CLR/CorLib/corlib_native_System_Convert.cpp
```diff
@@ -34,34 +37,45 @@ HRESULT Library_corlib_native_System_Convert::NativeToInt64___STATIC__I8__STRING
...
-            {
-                NANOCLR_SET_AND_LEAVE(CLR_E_OUT_OF_RANGE);
-            }
+            negReturnExpected = true;
         }
 
         // convert via strtoll / strtoull
+        int error_code;
+
+#ifdef _REENT_ONLY // Have to use reentrant version of strtoll because lwip sets _REENT_ONLY to require all stdlib calls
```
edleno2, line 57: This is the fundamental issue with LWIP - LWIP enables "_REENT_ONLY" which in some math routines requires that the calls be made in a re-entrant fashion.  Note that the Convert methods are not doing any reentrant work and the normal thread safety of errno would have worked.  I had already tried defining errno as an extern int, but that doesn't work since errno is really just a macro/define that hides a function call that returns the last error code for the current thread (making it thread safe).  With the "extern int errno" the code would compile, but actually fails the unit tests on the hardware because the error number is NOT going into the value "errno".  This fix - passing a structure that can include the errno - is how you are supposed to make these calls if _REENT_ONLY is defined.

src/CLR/Core/CLR_RT_HeapBlock.cpp
```diff
@@ -893,7 +893,7 @@ HRESULT CLR_RT_HeapBlock::PerformUnboxing(const CLR_RT_TypeDef_Instance &cls)
 
     if (this->DataType() != DATATYPE_OBJECT)
     {
-        NANOCLR_SET_AND_LEAVE(CLR_E_WRONG_TYPE);
```
edleno2, line 896: I believe these to be safe changes - the CLR_E_WRONG_TYPE returns a generic Exception type, so any try/catch in C# code will work the same.  I could not find any instances of CPP code specifically looking for CLR_E_WRONG_TYPE - everything I could find was just setting the value.

src/CLR/Include/nf_errors_exceptions.h
```diff
@@ -68,103 +68,103 @@ typedef int HRESULT;
 // clang-format off
```
edleno2, line 68: I think these are all of the code style changes you have in branch PR1936.  I screwed up rebasing using that branch, but all of the changes did get pulled into my fork.  In addition, I noticed there were a few success cases at the bottom of the file that still had "old" alignment, so I changed them to be consistent with all of the alignment you cleaned up.

src/CLR/CorLib/corlib_native_System_Convert.cpp
josesimoes: 🤪 that's convoluted but glad that you've sorted it out!! Well done!! 💯 🥇 👏🏻

src/CLR/Core/CLR_RT_HeapBlock.cpp
josesimoes: OK. Let's take those. If any changes are required in generics, I'll deal with those.

src/CLR/CorLib/corlib_native_System_Convert.cpp
josesimoes: If that's a WIN32 only thing the included could be wrapped in a #ifdef for WIN32 only. 
But if it's working as it is, even better.

josesimoes (approved):
Awesome!!

PR #1976, opened 2021-07-06
.NET Core target for nanoCLR

romanmashta:
## Motivation and Context

- .NET Core Native Host and CLI tools for hosting, running and debugging nanoCLR applications.

## Types of changes

## Checklist:


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/nanoframework/nf-interpreter?pullRequest=1976) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/nanoframework/nf-interpreter?pullRequest=1976) before we can accept your contribution.<br/><br/>:x: romanmashta [sign now](https://cla.dotnetfoundation.org/nanoframework/nf-interpreter?pullRequest=1976)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/nanoframework/nf-interpreter?pullRequest=1976) it.</sub>

edleno2:
I'll pull this down today and give it a try.  I looked over the change - 66 files - not sure how much of that is duplicate with the win32 target that is already in place, maybe that's where that code came from?  So maybe the intention is to replace win32 target with this one?  Still having trouble wrapping my head around the use case - can this be used without changes to either Device Explorer or nanoFramewor.TestFramework - those are the only two "entry" points for connecting to a nanoCLR I think?

romanmashta:
>not sure how much of that is duplicate with the win32 target that is already in place, maybe that's where that code came from? 

You can see from this commit that I took the win32 code as it was. https://github.com/nanoframework/nf-interpreter/pull/1976/commits/cb33aaf74fcd5c8c01c84ddfed19ca593b0808c6
And made changes to those files which required. I think it worth probably to remove unchanged files from solution and just include them from win32 folder. It will make this target lighter. And will avoid duplication of code. At the moment I didnt know what files I will have to update. 

What do you think? Should I remove unchanged duplicates and include them from win32?

Thanks.

edleno2:
> You can see from this commit that I took the win32 code as it was. [cb33aaf](https://github.com/nanoframework/nf-interpreter/commit/cb33aaf74fcd5c8c01c84ddfed19ca593b0808c6)

I didn't look at the individual commits - just the overall changes.  I've got it on my machine right now, but having some trouble getting all the parts to compile since I'm having to change to x64.  

I don't think it's a question of "how much to copy from win32" but whether to copy at all.  Should this be done using either a define or a runtime setting so that the new code is interspersed within the current win32 solution?  That's two approaches (do it at compile time, or do it at runtime).

An alternative is to plan on retiring the win32 target and we change the test framework to invoke the CLI component and not the current win32 nanoCLR.exe - I'll mention @josesimoes  and @Ellerbach to get their opinions.

I do see that the CLI takes a COM port as an argument, so that answers how we would integrate - you start the CLI project up in Visual Studio with debugger attached and then go to another instance of VS and run your NF project and select the COM port you put in the CLI command line.  Do I have that right?

edleno2:
Have to leave for a while - but if we continue with the forked version of nanoCLR(win32) then I'll want to make a suggestion on the handling of the AdditionalInclude directories so we can build in both x86 and x64.  Even after fixing that I still am not getting a good build of the CPP parts, so have to look closer to see what I missed.

romanmashta:
> Have to leave for a while - but if we continue with the forked version of nanoCLR(win32) then I'll want to make a suggestion on the handling of the AdditionalInclude directories so we can build in both x86 and x64. Even after fixing that I still am not getting a good build of the CPP parts, so have to look closer to see what I missed.

All projects in solution have to be compiled for x86. Including .net ones. I was targeted x86 initially because the original win32 project was x86 only from the beginning. I will try to make it compile and work for x64 also.

edleno2:
Yes, I did switch to x86 and all projects built then.  The original nanoCLR was "Any CPU" and I would have expected it to build as x64 on my development machine.  And I just glanced at the original nanoCLR and there weren't any additional includes in the project file, so we might want to look at why there are additional include in the new solution.  Additional Includes tend to be a problem because they can cause problems when developers lay their solutions out a bit differently on their machine.  Not sure that's the case here, but I avoid anything that can break on another developers machine.

edleno2:
I kind of forgot that you converted nanoCLR from an application to a library, so that explains more of your need to clone code.  We just have to decide on what to do with the current test framework and then figure out where to take this from there.  If we change to have the test framework use the CLI then we will have to copy the CLI .exe and the necessary .dll's during the post build of any test solution.  Not the end of the world, just a necessary change if we decide to go that way and kill off the win32 target that we have today.

josesimoes:
@edleno2 @romanmashta let's continue this conversation on Discord, shall we? It's more productive than here... 😉

targets/netcore/nanoFramework.nanoCLR.Native/Various.cpp
```diff
@@ -0,0 +1,513 @@
...
+{
+    (void)level;
+    (void)wakeEvents;
+};
+
+///////////////////////////////////////////////////////////////
+
+int nanoBooter_GetTargetInfo(TargetInfo *targetInfo)
+{
+    targetInfo->BooterVersion.usMajor = 0;
```
josesimoes, line 490: Same here: it's better to change it in the common various.cpp

targets/netcore/nanoFramework.nanoCLR.CLI/Program.cs
```diff
@@ -0,0 +1,68 @@
...
+        static void Main(string[] args)
+        {
+            LogErrors(() =>
+            {
+                NanoClrHostBuilder hostBuilder = NanoClrHost.CreateBuilder();
+                hostBuilder.UseConsoleDebugPrint();
+
+                Parser.Default.ParseArguments<CommandLineOptions>(args).WithParsed(o =>
+                {
+                    if (o.Assemblies.Any())
```
josesimoes, line 24: As per the project code style, please use open/close brackets, even if there a single line of code.
(multiple places with this)

josesimoes (changes requested):

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Text.RegularExpressions;
+using nanoFramework.nanoCLR.Host.Interop;
+using nanoFramework.nanoCLR.Host.Port;
+using nanoFramework.nanoCLR.Host.Port.NamedPipe;
+using nanoFramework.nanoCLR.Host.Port.Serial;
+using nanoFramework.nanoCLR.Host.Port.TcpIp;
+
+namespace nanoFramework.nanoCLR.Host
+{
+    public class NanoClrHostBuilder
```
Ellerbach, line 19: It's a great practice to document all public classes, there public methods and properties.
It helps maintaining the code.
Would be great if you can add. There are quite some. Leaving the comment here but it's valid for quite some classes.
Those classes will be extracted and placed in the static documentation website.

targets/netcore/nanoFramework.nanoCLR.Host/Port/Channel.cs
```diff
@@ -0,0 +1,62 @@
...
+                PortB.Process();
+                if (PortB.BytesAvailable > 0)
+                {
+                    PortA.TransmitData(PortB.ReceiveData());
+                }
+
+                PortA.Process();
+                if (PortA.BytesAvailable == 0 && PortB.BytesAvailable == 0)
+                {
+                    Thread.Sleep(100);
```
Ellerbach, line 57: Why 100 and not 50 or 10 of 300?
Would be perfect to add a comment so it will make it easy to maintain.

targets/netcore/nanoFramework.nanoCLR.Host/Port/Common/ListeningPortBase.cs
```diff
@@ -0,0 +1,69 @@
...
+//
+
+using System;
+using System.Threading;
+
+namespace nanoFramework.nanoCLR.Host.Port.Common
+{
+    public abstract class ListeningPortBase : IPort
+    {
+        private const int WaitTimeout = 100;
```
Ellerbach, line 13: Why 100 and not 50 or 10 of 300?
Would be perfect to add a comment so it will make it easy to maintain.

targets/netcore/nanoFramework.nanoCLR.Host/Port/NamedPipe/NamedPipeListeningPort.cs
```diff
@@ -0,0 +1,47 @@
...
+// See LICENSE file in the project root for full license information.
+//
+
+using System.IO;
+using System.IO.Pipes;
+using nanoFramework.nanoCLR.Host.Port.Common;
+
+namespace nanoFramework.nanoCLR.Host.Port.NamedPipe
+{
+    public class NamedPipeServerPort : ListeningPortBase
```
Ellerbach, line 12: Stupid question: I though we would get out of named pipe? So is this used somewhere else?

targets/netcore/nanoFramework.nanoCLR.Host/Port/TcpIp/TcpIpListeningPort.cs
```diff
@@ -0,0 +1,40 @@
...
+using nanoFramework.nanoCLR.Host.Port.Common;
+
+namespace nanoFramework.nanoCLR.Host.Port.TcpIp
+{
+    public class TcpIpListeningPort : ListeningPortBase
+    {
+        private readonly TcpListener _server;
+        private const string LocalHost = "127.0.0.1";
+
+        public TcpIpListeningPort(int port)
```
Ellerbach, line 17: Will there be a way to listen over a real network, not just locally? Maybe for later on?

Ellerbach (commented):
Few comments, all up, looks good! Impatient to test it!

targets/netcore/nanoFramework.nanoCLR.Host/Port/NamedPipe/NamedPipeListeningPort.cs
romanmashta: This Pull Request was made before our call. I've tried to implement it also. Will remove it.

targets/netcore/nanoFramework.nanoCLR.Host/Port/TcpIp/TcpIpListeningPort.cs
romanmashta: Sure, it can listen on any interface. I will change it to 0.0.0.0 as default and will add an extra parameter if user needs to specify exact IP to bind to.

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
romanmashta: Yes, I will try to add comments.

targets/netcore/nanoFramework.nanoCLR.Host/Port/Channel.cs
romanmashta: I've reimplemented Wire ports to use Async Receive and Transmit, so these loops are not required anymore.

targets/netcore/nanoFramework.nanoCLR.Host/Port/Common/ListeningPortBase.cs
romanmashta: I've reimplemented Wire ports to use Async Receive and Transmit, so these loops are not required anymore.

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
romanmashta: I've added documentation comments to public classes and methods

Ellerbach (approved):
Thanks! Looks very good to me!
Question: why is the vspt.vsptlic file empty?

romanmashta:
@Ellerbach 
> Question: why is the vspt.vsptlic file empty?

I had concerns if it is okay to put valid license file on Github. Need to get clarification from Jose on that. It could be possible to replace valid license file on npm release build for distribution, but I'm not sure if it is okay to share it on GitHub. Also, it is not a blocker for local builds and tests. Virtual com management will run in the trial 14 days mode when the license file is empty.

josesimoes (approved):

PR #1989, opened 2021-07-19
Remove need for source control in chibios

networkfusion:
when using local source versions of libs.

## Motivation and Context
When using local source folders for chibios (or libs) it currently requires the folders to already be initialized by Git or SVN.
This PR removes that need.

TODO: (if required to make it better):
* As a fallback, download the sources from the latest release bin,
* Remove the linux (dev container) difference.

## How Has This Been Tested?
Tested using a local source folder of chibios on an OrgPal3.

## Types of changes

## Checklist:

### DIFF
--- a/targets/ChibiOS/CMakeLists.txt
+++ b/targets/ChibiOS/CMakeLists.txt
@@ -92,16 +92,14 @@ else()
     # ChibiOS source was specified
 
     # sanity check is source path exists
-    if(EXISTS ${RTOS_SOURCE_FOLDER}/)
+    if(EXISTS ${RTOS_SOURCE_FOLDER}/) # TODO: should also check content!
         if(CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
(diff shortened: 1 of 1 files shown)


networkfusion:
Of note, this seems to cause issues on second build:
image

josesimoes (commented):
The check for content is noticiable slow and adds significantly to the time of local builds.

The fetch content call takes care of checking the content and pull changes from upstream, if there are any. By using the appropriate tool (svn or git) that happens effortlessly.
I'm aware of the issues with not having a got repo, but we have to deal with that the best possible way. All that we have now is trying to address the pipeline VS local VS container builds in a way that's not optimal for all of those...

networkfusion:
Hey @josesimoes , Surely this only happens if you add a local folder to the cmake-varients.json file, so in most circumstances for local build, the file will still be downloaded and cached via SVN/GIT (although we need to add something to detect that the client is installed). I think that is a worthy comprimise...

josesimoes:
@networkfusion if you are OK with this, let's have it merged.

josesimoes (approved):

networkfusion:
@josesimoes I am fine with it.

PR #2022, opened 2021-08-30
PSRAM pins in ESP32 are now reserved only if needed

Ellerbach:
## Description

Removing reserved pins setup on ESP32 for SPI flash in the GPIO controller.

Rational: if SPI Flash is used, then some of the pins will be used by SPI anyway. Not all the pins will be used all the time. The design of some of the ESP32 board (like M5 Stick) uses pins 9 and 10 as they are not used in SPI Flash while the board has some. And those 2 pins are used for IR and Red led.

## Motivation and Context

- Fixes https://github.com/nanoframework/Home/issues/813

## Types of changes

## Checklist:

### DIFF
--- a/targets/FreeRTOS_ESP32/ESP32_WROOM_32/nanoCLR/Windows.Devices.Gpio/cpu_gpio.cpp
+++ b/targets/FreeRTOS_ESP32/ESP32_WROOM_32/nanoCLR/Windows.Devices.Gpio/cpu_gpio.cpp
@@ -130,6 +130,16 @@ bool CPU_GPIO_Initialize()
     // Make sure all pins are not reserved
     memset(pinReserved, 0, sizeof(pinReserved));
 
+    // check if PSRAM it's available (querying largets free block available with SPIRAM capabilities)
+    if (heap_caps_get_largest_free_block(MALLOC_CAP_8BIT | MALLOC_CAP_32BIT | MALLOC_CAP_SPIRAM))
+    {
+        // Reserve Pins 6-11 as used by Spi flash
+        for (int pinNumber = 6; pinNumber <= 11; pinNumber++)
+        {
+            CPU_GPIO_ReservePin(pinNumber, true);
+        }
+    }
+
     // Install ISR service for GPIO
     esp_err_t ret = gpio_install_isr_service(0);
     if (ret != ESP_OK)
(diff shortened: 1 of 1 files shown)


josesimoes (commented):
I would try to make this a bit "smarter" and fool proof...
Meaning that the pin reservation should be there when the SPI flash it's present.

Ellerbach:
> Meaning that the pin reservation should be there when the SPI flash it's present.

I honestly don't see a point of reserving pins. If they are used by SPI flash, you won't anyway be able to open them, you'll get an error. So reserving pins that anyway will return an error if you want to open them doesn't feel logical.

josesimoes:
> > Meaning that the pin reservation should be there when the SPI flash it's present.
> 
> I honestly don't see a point of reserving pins. If they are used by SPI flash, you won't anyway be able to open them, you'll get an error. So reserving pins that anyway will return an error if you want to open them doesn't feel logical.

🤔 I don't think that's the case... you can reassign the GPIO pins use by SPI flash without getting any exception. The result is that, inadvertently, a developer can screw the execution and make the CLR unusable. Having this check layer was one of the motivations for adding all this CPU pins layer code.

@AdrianSoundy your input here, please.

AdrianSoundy:
Yes the problem is you don't get an error when pins are used that shouldn't be used.  The reserved pins was to make sure that doesn't happen and developer gets a clear message the pin was used elsewhere. 
For this case we just need to remove reserved pins for the flash pins if there is no PSRAM.

Could set a flag in memory.cpp or use :-
size_t spiramMaxSize = heap_caps_get_largest_free_block(MALLOC_CAP_8BIT | MALLOC_CAP_32BIT | MALLOC_CAP_SPIRAM);
To test if there is any psram memory available.

josesimoes:
Precisely! That check should be all it takes to reserve those pins **only** when they are being used by PSRAM. 👍🏻

Ellerbach:
> Precisely! That check should be all it takes to reserve those pins **only** when they are being used by PSRAM. 👍🏻

I would argue that you always have the board pin availability. So as a developer, you can only pins that are physically accessible. If they are used by the psram, they are not physically accessible, so you 1. have no reason to use them, 2. if ever you want to use them, when trying to open them, you'll get an error.

So I really don't think that reserving them is less confusing than not reserving them as anyway, you'll get an error at the end of the day.

Ellerbach:
So I think I finally got it. The detection of PSRAM is done at boot time, before this code is executed. So at this stage, it's about understanding if there is some. And if there is some reserve the pins otherwise, just skip!

josesimoes:
@Ellerbach this is building OK now. Care to check if it's working on your test app?

Ellerbach:
> @Ellerbach this is building OK now. Care to check if it's working on your test app?

any already built nanoCLR I can just download and test? (otherwise, will pull and build locally)

josesimoes:
> > @Ellerbach this is building OK now. Care to check if it's working on your test app?
> 
> any already built nanoCLR I can just download and test? (otherwise, will pull and build locally)

Sure! Just follow the URL to Azure Pipeline build and grab the bin from there. (nanoFramework it's developer friendly) 😉

Ellerbach:
Works like a charm on en ESP32 PICO (M5 Stick) blinking the led on GPIO 10. Congratulations!

PR #2050, opened 2021-09-19
Skip CLR_DBG_Monitor_Message while reading payload

doingnz:
## Description
Add check and skip sending CLR Debugger Monitor_Message from nF if, and only if, it is busy receiving a payload of an incoming Wire Protocol packet. 
The change only impacts:
```
void WP_PrepareAndSendProtocolMessage(uint32_t cmd, uint32_t payloadSize, uint8_t *payload, uint32_t flags)
```

## Motivation and Context
The Monitor_Message is sent continuously approximately every 1.2 seconds from nF to the PC host after a Debugger has connected with the firmware. The message is transmitted regardless of what else is going on. If by chance it is send when nF is in the middle of receiving a payload that is larger than the low-level hardware buffers, there is a risk the WP_Process loop reading the incoming payload will be delayed during the setup and transmission process of the Monitor Message and the low-level RX  buffer can overflow if it is small. Example buffer needs to be more than 300 bytes to avoid overflow. 

The problem is easiest to demonstrate if there is a deployed application (e.g. nBlinky) currently being executed by nF, when at the same time one is trying to send large payloads via WP.

I have not been able to reproduce the problem when the deployment area is erased. 

## How Has This Been Tested?
Build nF with the following settings:
mcuconf.h :
```
#define STM32_SERIAL_USE_UART4              TRUE
#define STM32_UART_USE_UART4                FALSE
```
with cmake-variants.json enabled Stdio trace on for errors. 
```
                    "NF_TRACE_TO_STDIO": "ON",
                    "NF_WP_TRACE_ERRORS": "ON",
```
Deploy nBlinky and leave it running. 
Connect with modified nanoFramework.Tools.Debugger test application with button that performs RAM "ERASE" (memset to 0xFF), Read & Write WP operations of 1024 bytes in a loop. Read used to check "Erase" and Write successful. [Used RAM for testing to avoid harm to flash with WR/ERASE.]

## Types of changes

## Checklist:

### DIFF
--- a/src/CLR/WireProtocol/WireProtocol_Message.c
+++ b/src/CLR/WireProtocol/WireProtocol_Message.c
@@ -425,6 +425,13 @@ void WP_PrepareAndSendProtocolMessage(uint32_t cmd, uint32_t payloadSize, uint8_
 {
     WP_Message message;
 
+    if (_rxState == ReceiveState_ReadingPayload && cmd == CLR_DBG_Commands_c_Monitor_Message)
+    {
+        // skip the outgoing Monitor message while we are busy receiving a packet payload
+        TRACE0(TRACE_ERRORS, "Skip CLR_DBG_Commands_c_Monitor_Message\n");
+        return;
+    }
+
     WP_Message_Initialize(&message);
 
     WP_Message_PrepareRequest(&message, cmd, flags, payloadSize, payload);


josesimoes:
I understand what you're trying to accomplish with this and it make sense.
Now, I think that we should address this at another level. The "offening

josesimoes:
I understand what you're trying to accomplish and it makes sense. Now, I think we should address it in a more generic way and without dropping any message.

If we add that check for _rxState == ReceiveState_ReadingPayload at the begining of WP_TransmitMessage it will become more generic and prevent that "overrun" on any situation. 
Ideally we would have a semaphore but, to make it simpler, we can have a call to a sleep for N ms (the same as the usual timeout for RX) and then check again a proceed to the TX or exit with error, if not.

What do you think about this?

doingnz:
Most [al?]l other WP TX/RX data exchanges are in response to the PC Debugger asking the nF Debugger to do something. The PC Debugger is waiting for a response and will not send another WP till it's previous request has been responded to or a timeout happens. 

The CLR_DBG_c_Monitor_Message is different to all the other WP TX from nF as it is autonomously initiated from nF Debugger and it is ONLY active after an initial connect from a PC Debugger. I have not looked in detail, but assume there is a timer event that is triggering this transmit a 1.2s intervals to tell the PC Debugger the nF Debugger is still attached and active.

I expect that we don't need to worry about any of the other WP TX from nF as they will not happen in parallel with receiving a large payload. The only large payloads coming into nF Debugger are FLASH/RAM WR operations? All other payloads I expect are small and mostly just 32 bytes of header and small number of bytes of payload that can easily fit in the low level serial IRQ buffer if it is 64 bytes. 

The true fix is to move to DMA based handling of the WP. DMA will work nicely given the headers are fixed size and include payload size... the interesting part of this work is the initial synchronisation to get to packet boundaries and recovery when there are errors.

I would recommend to not move this "hack" into lower level WP TX method unless we can show there is a need for it. Another approach is to create a mutex/semaphore exclusively used by the timer thread that initiates the  CLR_DBG_c_Monitor_Message send request. That would let the  CLR_DBG_c_Monitor_Message be sent directly after the incoming payload has been received.

doingnz:
Closing the PR. 

The edit is simple if we ever needed it again. I am not sure why there were erase/write issues during deployment a couple of days ago as the application should not be running..... 

I have seen some strange behavior in the PC Debugger test application at startup that is related to the way the async is running... it's as if there are multiple threads trying at startup to connect ... 

Problem appears to self resolve after a short interval when the "spare" threads error and disappear.  I can then run the WP stress test for ages and it's Ok.... something to comeback to later ...

PR #2060, opened 2021-09-20
Fix PWM deviceSelectorString

doingnz:
## Description
Insufficient space allocated when char deviceSelectorString[30] #ifndef STM32F0XX is true as there will be 7 timer strings "TIM1,TIM2,TIM3,TIM4,TIM5,TIM8,TIM9", each 5 char long. Current allocation is 30 when 35 is required. 

I suggest we let the compiler figure out the maths and suggest  char deviceSelectorString[7 * 5 + 1] = { 0 };

## Motivation and Context
Fix bug.

## How Has This Been Tested?
Source code review. 

## Types of changes

## Checklist:

### DIFF
--- a/targets/ChibiOS/_nanoCLR/Windows.Devices.Pwm/win_dev_pwm_native_Windows_Devices_Pwm_PwmController.cpp
+++ b/targets/ChibiOS/_nanoCLR/Windows.Devices.Pwm/win_dev_pwm_native_Windows_Devices_Pwm_PwmController.cpp
@@ -125,7 +125,7 @@ HRESULT Library_win_dev_pwm_native_Windows_Devices_Pwm_PwmController::GetDeviceS
     NANOCLR_HEADER();
     {
     // declare the device selector string whose max size is "TIM1,TIM3,TIM4,TIM5,TIM8,TIM9," + terminator and init with the terminator
-       char deviceSelectorString[ 30 + 1] = { 0 };
+       char deviceSelectorString[7 * 5 + 1] = { 0 };
 
    #if STM32_PWM_USE_TIM1
        strcat(deviceSelectorString, "TIM1,");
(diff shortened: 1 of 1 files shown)


doingnz:
I edited in GitHub for quick fix. 

Someone should review all the other places similar method used to calc list of devices.

An alternative approach is let the compiler concat the strings and avoid the overhead and bugs at runtime.  
```
    static char deviceSelectorString[] = ""
#if STM32_PWM_USE_TIM1
                                         + "TIM1,"
#endif
#if STM32_PWM_USE_TIM2
                                         + "TIM2,"
#endif
#if STM32_PWM_USE_TIM3
                                         + "TIM3,"
#endif
#ifndef STM32F0XX
#if STM32_PWM_USE_TIM4
                                         + "TIM4,"
#endif
#if STM32_PWM_USE_TIM5
                                         + "TIM5,"
#endif
#if STM32_PWM_USE_TIM8
                                         + "TIM8,"
#endif
#if STM32_PWM_USE_TIM9
                                         + "TIM9,"
#endif
#endif
        ;
```

Yet another alternative is define the string in a target file as is used for other hw devices like I2C, SPI etc... 
e.g. in a file targets\ChibiOS\NETDUINO3_WIFI\target_windows_devices_pwm_config.h

josesimoes:
> An alternative approach is let the compiler concat the strings and avoid the overhead and bugs at runtime.
> 

I like this approach better. It puts the burden on the machine. Less error prone. 😉 
Can you please change it to this one?

doingnz:
I agree the machine should do the work. 

We can then also get rid of the #ifndef STM32F0XX/#endif as the selected timers are in mcuconf.h

doingnz:
The list of possible timers on some devices is much more than currently listed here.

I have left list as is to simply fix the current bug.

josesimoes (approved):
Nice catch! Thanks for the fix.

PR #2080, opened 2021-09-28
Improve S.IO.Ports

josesimoes:
## Description
- Update native code to handle most serial tasks in native code.
- Update S.IO.Ports assembly declaration.

## Motivation and Context
- General performance improvement.
- Following nanoframework/System.IO.Ports#32.

## Types of changes

## Checklist:

### DIFF
--- a/targets/ChibiOS/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
+++ b/targets/ChibiOS/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
@@ -36,6 +36,56 @@
 // NF_PAL_UART Uart8_PAL;
 // #endif
 
+static NF_PAL_UART *GetUartPAL(int index)
+{
+    // Choose the driver for this SerialDevice
+    switch (index)
+    {
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_USART1) && (NF_SERIAL_COMM_STM32_UART_USE_USART1 == TRUE)
+        case 1:
+            return &Uart1_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_USART2) && (NF_SERIAL_COMM_STM32_UART_USE_USART2 == TRUE)
+        case 2:
+            return &Uart2_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_USART3) && (NF_SERIAL_COMM_STM32_UART_USE_USART3 == TRUE)
+        case 3:
+            return &Uart3_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_UART4) && (NF_SERIAL_COMM_STM32_UART_USE_UART4 == TRUE)
+        case 4:
+            return &Uart4_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_UART5) && (NF_SERIAL_COMM_STM32_UART_USE_UART5 == TRUE)
+        case 5:
+            return &Uart5_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_USART6) && (NF_SERIAL_COMM_STM32_UART_USE_USART6 == TRUE)
+        case 6:
+            return &Uart6_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_UART7) && (NF_SERIAL_COMM_STM32_UART_USE_UART7 == TRUE)
+        case 7:
+            return &Uart7_PAL;
+#endif
+
+#if defined(NF_SERIAL_COMM_STM32_UART_USE_UART8) && (NF_SERIAL_COMM_STM32_UART_USE_UART8 == TRUE)
+        case 8:
+            return &Uart8_PAL;
+#endif
+        default:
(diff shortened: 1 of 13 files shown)


targets/ChibiOS/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
```diff
@@ -276,816 +286,844 @@ HRESULT Library_sys_io_ser_native_System_IO_Ports_SerialPort::get_BytesToRead___
...
+        }
 
-    // Check RS485 mode is not selected as currently not supported
-    if ((SerialMode)pThis[FIELD___mode].NumericByRef().s4 != SerialMode_Normal)
+        // fill data buffer from Rx buffer
+        palUart->RxRingBuffer.Pop(buffer, bufferLength);
+
+        NANOCLR_CHECK_HRESULT(CLR_RT_HeapBlock_String::CreateInstance(top, (const char *)buffer, bufferLength));
+
+        platform_free(buffer);
```
Ellerbach, line 475: You have a risk here that the buffer won't be free if the previous check fails.

targets/ChibiOS/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
```diff
@@ -276,816 +286,844 @@ HRESULT Library_sys_io_ser_native_System_IO_Ports_SerialPort::get_BytesToRead___
...
+    // pop "hbTimeout" heap block from stack
+    stack.PopValue();
+
+    // return how many bytes were read
+    stack.SetResult_String((const char *)line);
+
+    // free memory, if needed
+    if (line != NULL)
+    {
+        platform_free(line);
```
Ellerbach, line 574: Same risk here as you have checks before. So just raising this to make sure it's not a possibility and that the line will be freed whatever happens before

targets/ChibiOS/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
```diff
@@ -276,816 +286,844 @@ HRESULT Library_sys_io_ser_native_System_IO_Ports_SerialPort::get_BytesToRead___
...
+
+    // pop "hbTimeout" heap block from stack
+    stack.PopValue();
+
+    stack.SetResult_U4(length);
+
+    // free memory, if it was allocated
+    if (isNewAllocation && buffer)
+    {
+        platform_free(buffer);
```
Ellerbach, line 1122: And same question about the risk

targets/FreeRTOS/NXP/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
```diff
@@ -177,536 +182,656 @@ HRESULT Library_sys_io_ser_native_System_IO_Ports_SerialPort::get_BytesToRead___
...
-        }
+    // return how many bytes were read
+    stack.SetResult_String((const char *)line);
 
-        // null pointers and vars
-        pThis = NULL;
+    // free memory, if needed
+    if (line != NULL)
+    {
+        platform_free(line);
```
Ellerbach, line 481: And one more similar question here as well

targets/FreeRTOS/NXP/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
```diff
@@ -177,536 +182,656 @@ HRESULT Library_sys_io_ser_native_System_IO_Ports_SerialPort::get_BytesToRead___
...
-        base->CTRL |= 1U << LPUART_CTRL_ORIE_SHIFT;
-        // Renable transmitter and receiver
-        base->CTRL |= 1U << 19;
-        base->CTRL |= 1U << 18;
+        // fill data buffer from Rx buffer
+        palUart->RxRingBuffer.Pop(buffer, bufferLength);
+
+        NANOCLR_CHECK_HRESULT(CLR_RT_HeapBlock_String::CreateInstance(top, (const char *)buffer, bufferLength));
+
+        platform_free(buffer);
```
Ellerbach, line 377: I guess I won't ask the question everywhere :-) There are something like 15+ of those :-) Just want to make sure whatever happens the buffer is properly freed

Ellerbach (approved):
WOW. Amazing work and refactoring. Very impressive. Only questions I have is regarding the platform_free, just to make sure they always happen. Reason is because there is some code checking for HRESULT and redirecting to the end of the function. And if the bugger has been allocated before and for some reason that check will fail, then it will never be deallocated.
Other than that, rest looks excellent!

targets/ChibiOS/_nanoCLR/System.IO.Ports/sys_io_ser_native_System_IO_Ports_SerialPort.cpp
josesimoes: Good catch on this one! I did try to make those calls to `platform_free` to always be on the exit path. 😅 
Sometimes it does look that's been missed, but that's when the buffer hasn't been allocated.
I'll double check these again and add comments for those cases.

PR #2083, opened 2021-09-30
Simpler top-level cmakelists

doingnz:
## Description
Removed most of the custom RTOS & target logic as the approach encouraged additional script for each additional target. 
This PR does not change the logic for the two ESP targets other than to use the same named CMake variables for the RTOS and Target folders to show how close it is to being the same as the other targets. Given the ESP targets are being reworked for the update to 4.x, I have left that change to the person doing the ESP update to remove ESP difference.

## Motivation and Context
Ideally there would be no RTOS or TARGET specific logic in this top-level CMakeLists.txt to aim for reduced responsibility and remove chance of unexpected interactions when new platforms added. its purpose should be restricted to check a target exists and validate high level shared options common to all nanoFramework targets. It also manages the selection of the standard target or community target. 

With this change, a new ${RTOS} & ${TARGET} will be found simply by creation of the folders under /targets. i.e. /targets/${RTOS}/${TARGET}

## How Has This Been Tested?
Confirmed can build ESP32 and STM32 targets. 

## Types of changes

## Checklist:


CMakeLists.txt
```diff
@@ -64,51 +64,16 @@ if(NOT RTOS OR RTOS STREQUAL "")
...
-set(RTOS_SUPPORTED "ChibiOS" "FreeRTOS" "FreeRTOS_ESP32" "TI_SimpleLink" CACHE INTERNAL "supported RTOS options")
+# check the RTOS and TARGET folders exist. 
+# optionally add custom logic in the last section of this CMakeLists.txt 
+# to perform custom add_subdirectory() actions.
 
-list(FIND RTOS_SUPPORTED ${RTOS} RTOS_INDEX)
-if(RTOS_INDEX EQUAL -1)
-    message(FATAL_ERROR "\n\n'${RTOS}' is an invalid option for RTOS. Please choose a valid RTOS.\n\n")
+if(EXISTS ${CMAKE_SOURCE_DIR}/targets/${RTOS})
+    message(STATUS "\nSetting RTOS to ${RTOS} Exists!\n")
```
josesimoes, line 72:
```suggestion
    message(STATUS "\nSetting RTOS to ${RTOS}\n")
```

CMakeLists.txt
```diff
@@ -774,45 +739,13 @@ endif()
...
-        endif()
-
-    endif()
+# Define PLATFORM base path
+set(BASE_PATH_FOR_PLATFORM ${CMAKE_CURRENT_SOURCE_DIR})
 
 #######################
 # FreeRTOS
-elseif(RTOS_FREERTOS_CHECK)
+if(RTOS_FREERTOS_CHECK)
```
josesimoes, line 748: I guess this should be:
```suggestion
if(RTOS_FREERTOS_ESP32_CHECK)
```

josesimoes (changes requested):

CMakeLists.txt
doingnz: There are two platforms at the start of the "custom" platforms logic related to FreeRTOS. Please see a couple of lines below your suggested edit the if(RTOS_FREERTOS_ESP32_CHECK)

```
#######################
# FreeRTOS
if(RTOS_FREERTOS_CHECK)

    add_subdirectory(targets/FreeRTOS)

#######################
# FreeRTOS ESP32
elseif(RTOS_FREERTOS_ESP32_CHECK)
```

josesimoes: If this is the IF for FREERTOS, then the code below doesn't belong here as it's validating RTC for ESP32, which would belong to _another_ IF....

josesimoes: Forget my comment above. I was mislead by the partial representatiuon of the diff on git. After looking at the complete file, I can see the code block is correct.

josesimoes (approved):
LGTM!

nightlark:
@josesimoes I'm not sure how nfbot is deciding when to apply the `invalid` label, since it seems to be getting added to merged PRs and ones that have been closed and replaced by a newer PR. Given that hacktoberfest is apparently counting `invalid` labels towards disqualification, use of the label by nfbot should probably be paused until after October.

josesimoes:
> @josesimoes I'm not sure how nfbot is deciding when to apply the `invalid` label, since it seems to be getting added to merged PRs and ones that have been closed and replaced by a newer PR. Given that hacktoberfest is apparently counting `invalid` labels towards disqualification, use of the label by nfbot should probably be paused until after October.

@nightlark criteria for adding invalid label is for PRs that where closed without being merged. This one must have been a glitch resulting from the order the git events where processed. The code processing these hasn't been changed for months now. I've also randomly checked several other PRs and couldn't find any that was wrongly labeled. 
Thanks for sharing your concern! 😃 👍🏻

PR #2091, opened 2021-10-03
Add networking to AzureRTOS with ChibiOS HAL - Draft

doingnz:
## Description
Initial steps towards adding networking to AzureRTOS and specifically the ST_B_L475E_IOT01A target. 

## Motivation and Context
Adds new feature of networking using the Wifi module on the ST_B_L475E_IOT01A

## How Has This Been Tested?
Work is incomplete as the wifi driver needs modification to use ChibiOS SPI HAL

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


CMake/Modules/FindAzureRTOS.cmake
```diff
@@ -9,7 +9,7 @@
 
 # check if the series name is supported 
 
-set(AZURERTOS_SUPPORTED_SERIES STM32L4xx RP2040 MICROBIT MAX78000 CACHE INTERNAL "supported series names for AzureRTOS")
+set(AZURERTOS_SUPPORTED_SERIES STM32L4xx RP2040 MICROBIT MAX78000 CACHE INTERNAL "supported series names for ${RTOS}")
```
josesimoes, line 12: These CMakes are AzureRTOS specific, therefore I don't see any value on replacing the name string with a variable that has to be processed by CMake for replacement of that same string. Seems a waste of processing to me...
(this comment it's valid for all the similar ones found ahead)

CMake/Modules/FindChibiOS_F0_HAL.cmake
```diff
@@ -80,8 +80,8 @@ list(APPEND ChibiOSnfOverlay_SOURCES ${CMAKE_SOURCE_DIR}/targets/ChibiOS/_nf-ove
 # adjust search path here
 if(RTOS_CHIBIOS_CHECK)
     set(OSHAL_PATH ${chibios_SOURCE_DIR}/os/hal/osal/rt-nil)
-elseif(RTOS_AZURERTOS_CHECK)
-    set(OSHAL_PATH ${CMAKE_SOURCE_DIR}/targets/AzureRTOS/ChibiOS/HAL)
+elseif(RTOS_AZURERTOS_CHIBIOS_HAL_CHECK)
```
josesimoes, line 83: I don't see the need to add _another_ variable for AzureRTOS + ChibiOS...
Even if we have AzureRTOS + _ACME_ HAL this would be including it's own HAL CMake, not this one. In the end this check here would be redundant or not bringing much value to the build.

Same comment is valid for the other ChibiOS HAL files that follow.

CMake/binutils.AzureRTOS.cmake
```diff
@@ -253,9 +255,14 @@ macro(nf_add_platform_include_directories target)
     # includes specific to nanoCRL
     if(${target} STREQUAL ${NANOCLR_PROJECT_NAME})
 
+        message(STATUS "# includes specific to nanoCRL ++++++++++++++++")
```
josesimoes, line 258: These seem to be outputs helpful for debugging. Please either comment out for the time being or wrap with `if (BUILD_VERBOSE)` if they are really important and deserve to be here.

CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/CMake/Modules)
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/CMake)
 ########################################################
 
+unset(NANOCLR_PROJECT_INCLUDE_DIRS CACHE)
```
josesimoes, line 36: Why do we need these here?

targets/AzureRTOS/ST/CMakeLists.txt
```diff
@@ -84,13 +84,20 @@ endif()
 FetchContent_GetProperties(chibios)
 FetchContent_Populate(chibios)
 
-list(APPEND TARGET_AZURERTOS_COMMON_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/include)
+list(APPEND TARGET_AZURERTOS_COMMON_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/_include)
```
josesimoes, line 87: Why dropping the `_` in the include path?

josesimoes (changes requested):
Richard: please let's keep the commits and PR atomic. 
This one it's mixing an improvement in the Azure RTOS specific by moving blocks from main CMake to platform one, that's OK. But then it's mixing network bits and HAL. I can see value in the 1st part. Let's have a PR for that only. Then focus on the network on this one.

Also you're adding netxduo source to our repo. This goes against a decision that was made at the beginning of the project and would like to stick with it. Netxduo sources can (and have to) be added to the build as a separate component like we're doing for all the other components.

CMake/binutils.AzureRTOS.cmake
doingnz: this is a draft PR ... they would not be in the final squashed commit

targets/AzureRTOS/ST/CMakeLists.txt
doingnz: the folder relative to the ST folder is called '_include' not 'include'... we can rename the folder if you prefer without the '_'

CMakeLists.txt
doingnz: Initially did not want to add utilities.cmake to the main CMake folder as we may not want it long term. I also wanted to discuss a concern about adding platform specific cmake files in a shared location. Makes it difficult to have something similar for another platforms.

doingnz:
a limited and small portion of netxduo source was added as per the original sample from Microsoft for this initial platform.

targets/AzureRTOS/ST/CMakeLists.txt
josesimoes: I read it the other way around. The way to go is with the _ prefix. So we're good on this.

PR #2170, opened 2021-12-09
Improve managed heap allocation

josesimoes:
## Description
- Rework code to simplify allocation of managed heap memory.
- Now always using  heap caps API to get largest free block.
- Simplification of definitions on how much memory to leave free.

## Motivation and Context
- Have managed heap allocation dynamic instead of fixed, according to device capabilities, by having a definition of the amount of RAM to leave free for malloc, instead of the opposite approach which was setting upfront the managed heap size.

## How Has This Been Tested?
- Tested with various images with and without BT. Run apps with high demand on internal memory, like SSL sample. Run BluetoothLESerial sample.

## Types of changes

## Checklist:


AdrianSoundy:
With this change we would go back to maximum available memory for managed heap.
This was previously an issue for TLS applications.  Although that was on IDF v3.3.5. Maybe something changed for v4.3.1 in the TLS area, Preallocation of buffers or IDF config ?

For BLE i had issues with initializing BLE stack until I reduced managed heap but that was also on v3.3.5 so that may also be different now. Will need to do further testing.

I'll look further into this over weekend.
@josesimoes

josesimoes:
Both those scenarios are covered with these changes. With BT there is an extra free ram that's left available for malloc.
With TLS the buffers allocation is set to the minimum possible upfront (and we can also set  some amount to be left for malloc if needed).

I've tested both scenarios as I've wrote in the PR comment. Both work without any noticeable issues)

targets/ESP32/_nanoCLR/Memory.cpp
```diff
@@ -93,73 +94,68 @@ void HeapLocation(unsigned char *&baseAddress, unsigned int &sizeInBytes)
...
-                ESP_LOGI(TAG, "Failed to allocate managed heap from SPIRAM");
-            }
+            ESP_LOGI(TAG, "Allocating managed heap from SPIRAM");
+        }
+        else
+        {
+            // allocate from internal RAM
+
+            // find out how much can be allocated?
+            largestFreeBlock = heap_caps_get_largest_free_block(MALLOC_CAP_8BIT | MALLOC_CAP_32BIT);
```
AdrianSoundy, line 132: Use memcaps
```suggestion
            largestFreeBlock = heap_caps_get_largest_free_block(memoryCaps);
```

AdrianSoundy (changes requested):
I have done some testing with this and heap_caps_get_largest_free_block is only returning a value of 65536 which is less then what we are currently  getting.  The problem is a IDF bug which i understand has been fixed or will be fixed.

see:- https://github.com/espressif/esp-idf/issues/7808

For managed heap we would get:-
ESP32_rev0 = 64K
Esp32_ble_rev0 = 44K
Which is why it works with test programs.

targets/ESP32/_nanoCLR/Memory.cpp
josesimoes: The fix for that bug has already been merged in the 4.3 branch. Make sure you have your local branch updated. On my tests I was getting >100k of heap at C# app start.

josesimoes: Good point. I've replaced it also on the other calls.

AdrianSoundy:
The PR itself is ok but there is a bug in heap_caps_get_largest_free_block that's only returning 65536. So if we merge this we will have less managed heap then we already have. 
I highlighted this in issue.  We can either wait for next IDF, 4.3.2 or patch IDF ourselves based on code suggested in IDF issue notes.

PR #2215, opened 2022-01-04
Improvements in CMAKE

networkfusion:
## Description
Use `IS_DIRECTORY` rather than `EXISTS`
Also fixes populate to make it possible to rebuild targets that use local source
AND removes the un-necessary linux specific SVN source for chibios.

## Motivation and Context
Improves check to make it consistant.

## Types of changes

## Checklist:


josesimoes:
Looks good! Any chance you can replace all the similar ones in other platforms  CMakes?

josesimoes:
@networkfusion I was experimenting with this and I think we can improve it even further!
The reason for those checks if the directory existed and it had content date back from the very beginning. Seems that the call to `FetchContent_Declare` performs those checks and outputs a clear enough message.

Here's the output when pointed to an empty directory or when the path doesn't exist:
```
[cmake] CMake Error at C:/Program Files/CMake/share/cmake-3.22/Modules/ExternalProject.cmake:2882 (message):
[cmake]   No download info given for 'ti_xdctools-populate' and its source directory:
[cmake] 
[cmake]    E:/GitHub/TI_XDCTools
[cmake] 
[cmake]   is not an existing non-empty directory.  Please specify one of:
[cmake] 
[cmake]    * SOURCE_DIR with an existing non-empty directory
[cmake]    * DOWNLOAD_COMMAND
[cmake]    * URL
[cmake]    * GIT_REPOSITORY
[cmake]    * SVN_REPOSITORY
[cmake]    * HG_REPOSITORY
[cmake]    * CVS_REPOSITORY and CVS_MODULE
``` 

I guess we can make that simplification and just remove that outer IF and make is even simpler.

robin-jones:
I also tried the `FetchContent_Declare` and ran into issues with CMAKE background building the folders source, which runs into errors in some of them, so might need to be careful there... But if you can make it work, I have nothing against it.

networkfusion:
@josesimoes , if I have done it correctly in the last commit (for TI) I will also go through the others...

targets/ChibiOS/CMakeLists.txt
```diff
@@ -416,24 +394,22 @@ if(NF_FEATURE_USE_SPIFFS)
...
-                GIT_TAG "nf-build"
-            )
+        message(STATUS "SPIFFS source from: ${SPIFFS_SOURCE}")
 
-        else()
-            message(FATAL_ERROR "Couldn't find SPIFFS source at ${SPIFFS_SOURCE}/")
-        endif()
+        FetchContent_Declare(
+            spiffs
+            GIT_REPOSITORY ${SPIFFS_SOURCE}
```
josesimoes, line 401: this should be SOURCE_DIR too

networkfusion: Done.

josesimoes (approved):
LGTM! Thanks.

PR #2308, opened 2022-05-09
Fix RingBuffer OptimizeSequence

AdrianSoundy:
## Description

There is an error in the tailsize handling for ringbuffer OptimizeSequence.
With a datasize of 1 and a Write_index=0 or 1 we end up trying to allocate -1 or 0 memory and tempBuffer will be NULL causing an exception. 

Done a quick change which i think works OK but probably can be optimized.

This caused a problem with System.io.ports

## Motivation and Context
Causing unhandled exceptions with System.IO.Ports 
 
## How Has This Been Tested?
Haven't had much time to test yet

## Types of changes

## Checklist


src/HAL/Include/nanoHAL.h
```diff
@@ -756,26 +756,33 @@ template <typename T> class HAL_RingBuffer
             // |xxxx......xxxxxx|
 
             // store size of tail
-            size_t tailSize = _write_index - (1 * _dataSize);
+            size_t tailSize = _write_index;
```
josesimoes, line 759: You're ignoring the data size on purpose? 🤔

AdrianSoundy: I don't think its required in this instance as write_index is just current byte index unlike where you push on elements or pop off elements.

I'll try an do some testing with different element sizes to confirm its working properly now.

AdrianSoundy: Seems to work for byte buffers but there are lots of issues with other sizes.
The read_index and Write_index are byte index but when ever it does a memcpy with a buffer the buffer calculations are wrong because its adds a byte index to a say Unit16_t buffer ptr.
We either need to put a (uint8_t *) to force the ptr types or always use a uint8_t buffer array.

josesimoes: I see... this was meant to be "generic" but probably hasn't been used with types other than byte...

Two options: either fix it to work properly with all sizes or rename and make it clear that's for bytes only.
Nothing against the latter.

AdrianSoundy: From testing Its working now for byte buffers ok but needs work for anything else.
Don't have time to work on this now so rather commit what i have and fix later for other data sizes (I'll create new issue for it)

Been just trying to track problems with my project. This is just one of the serial problems i have found.

AdrianSoundy: Let me know if you are happy with merging now.

AdrianSoundy: Added issue to follow up :- https://github.com/nanoframework/Home/issues/1042

josesimoes (approved):
Let's for it! 👍🏻

PR #2324, opened 2022-05-13
Fix hang in nanoClr when using serial with Hand shake is enabled

AdrianSoundy:
## Description

When you try to write small packet of data and the hand shake is stopped the driver will hang in the ESP32 IDF uart_write_bytes() call until flow is enabled again. Because of this all threads in managed code hang/stop.

Change done to to use different call to only write data to fifo and any further data is done using long running code path. Removed long running calculation at no longer needed.
 
In testing noticed that performance is noticeably faster as call is no longer waiting for serial to complete.

Also noticed when reviewing existing code that there is a potential of a memory leak with the writeString call when a new line is added to string and long running operation is used. (Not currentty fixed)

The easiest way to fix this is by moving the adding of new line to managed code so new buffer doesn't need to be allocated in native code. What do you thing?  

Otherwise would need to save buffer to remove in UART structure and free when operation complete.

## Motivation and Context
- Fixes nanoFramework/Home#1041
- 
## How Has This Been Tested?
Tested with serial simulator and now threads don't hang and when flow is enabled the serial data continues.

## Types of changes

## Checklist


AdrianSoundy:
I have tested this and it working well.
Before merging need to decide what to do about memory leak. 
1) Fix in serial ports by saving address to free in uart structure
2) Fix in managed code by adding new line before calling native code.

AdrianSoundy:
Why do we get 2 code style fixes ?
Both the same

josesimoes:
> Why do we get 2 code style fixes ? Both the same

Because there are 2 pipelines running and both have the code style check task.
Agreed that's a bit annoying but I haven't found a way of improving this...

Just merge one of them.

josesimoes:
> I have tested this and it working well. Before merging need to decide what to do about memory leak.
> 
> 1. Fix in serial ports by saving address to free in uart structure
> 2. Fix in managed code by adding new line before calling native code.

Looks like the simplest fix will be in C#, correct? Let's go for this one. One could argue about the performance impact, but I'm not seeing this being called in a pace that would be impacted by that. Not sure if you agree...

josesimoes (approved):
LGTM!

AdrianSoundy:
Ok i merge this and do a separate fix for memory leak. Its always been there so i am sure it can wait another day :-)

PR #2327, opened 2022-05-13
Cmake Presets improvements

networkfusion:
## Description
* Simplifies and improves the CmakePresets
* The changes were (trying to be) individual so that they can be cherry picked.

## Motivation and Context
* Some of the options were obsolete/duplicated or un-necessary.

## Types of changes

## Checklist


networkfusion:
!!! Ignoring code fix recommendations until the initial PR is checked over...

networkfusion:
SWO rename removed as caused more issues than it is worth (for this PR).

networkfusion:
Struggling with this one... 
image

All targets (as far as I can see) have the managed debugger enabled, but it is not the default option! Perhaps it should be changed, and a note about "RTM image..."

networkfusion:
No further changes will be made against this PR, as I think it is enough (unless review comments are made).

networkfusion:
No further changes will be made against this PR, as I think it is enough (unless review comments are made).

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -161,12 +161,30 @@ endmacro()
 
 # Hardware.Esp32
 if(API_Hardware.Esp32)
-    ##### API name here (doted name)
+    if(API_Hardware.Stm32)
```
josesimoes, line 164: I understand what you're trying to do with this checks.
The thing is that the build will break ALWAYS if any of those is selected with the wrong platform. Which make these redundant.

With all these new IF statements it just gets harder to maintain and raises the need for updating them all when adding a new platform.

Having said that: I don't think we need these here.

networkfusion: Ok, I reverted this commit so the file was not changed.

josesimoes (approved):
LGTM

PR #2340, opened 2022-05-23
Extends the SpiConfiguration to enable a manual chip-select line

KasperJSdeVries:
## Description
- Rework SPI PAL on ChibiOS, FreeRTOS and Azure RTOS to accept -1 as CS.
- Change ChibiOS HAL config to NOT handle CS line in the driver, which is not handled programatically. 
- Rework CPU SPI to properly handle CS being -1

## Motivation and Context
- Closes nanoFramework/Home#1045.

## Types of changes

## Checklist

### DIFF
--- a/src/HAL/Include/nanoHAL_Spi.h
+++ b/src/HAL/Include/nanoHAL_Spi.h
@@ -18,7 +18,7 @@ struct nanoSPI_BusConfig
 {
     bool spiBusInited;
     int8_t devicesInUse;
-    SPI_DEVICE_CONFIGURATION deviceCongfig[MAX_SPI_DEVICES];
+    SPI_DEVICE_CONFIGURATION deviceConfig[MAX_SPI_DEVICES];
     uint32_t deviceHandles[MAX_SPI_DEVICES];
     float byteTime[MAX_SPI_DEVICES];
     SPI_OP_STATUS spiStatus;
--- a/src/System.Device.Spi/nanoHAL_Spi.cpp
+++ b/src/System.Device.Spi/nanoHAL_Spi.cpp
@@ -279,20 +279,31 @@ HRESULT nanoSPI_OpenDeviceEx(
     }
 
     // Find if device slot is available and check
-    deviceIndex = FindFreeDeviceSlotSpi(spiDeviceConfig.Spi_Bus, spiDeviceConfig.DeviceChipSelect);
+    if (spiDeviceConfig.ManualChipSelect)
 
-    if (deviceIndex < 0)
     {
-        if (deviceIndex == -1)
-        {
-            // No device slots left
-            return CLR_E_INDEX_OUT_OF_RANGE;
-        }
-        else
-        {
-            // Return NOT_SUPPORTED when Device already in use. Not really any other relevant exception that's
-            // currently raised in managed code
+        if (spiconfig[spiBusIndex].devicesInUse > 0)
             return CLR_E_NOT_SUPPORTED;
+
+        deviceIndex = 0;
+    }
+    else
+    {
+        deviceIndex = FindFreeDeviceSlotSpi(spiBusIndex, spiDeviceConfig.DeviceChipSelect);
+
+        if (deviceIndex < 0)
+        {
+            if (deviceIndex == -1)
+            {
+                // No device slots left
+                return CLR_E_INDEX_OUT_OF_RANGE;
+            }
+            else
+            {
+                // Return NOT_SUPPORTED when Device already in use. Not really any other relevant exception that's
+                // currently raised in managed code
+                return CLR_E_NOT_SUPPORTED;
+            }
         }
     }
 
@@ -305,7 +316,8 @@ HRESULT nanoSPI_OpenDeviceEx(
         spiconfig[spiDeviceConfig.Spi_Bus].devicesInUse--;
 
         // Unreserve CS pin
-        CPU_GPIO_ReservePin(spiconfig[spiDeviceConfig.Spi_Bus].deviceCongfig->DeviceChipSelect, false);
+        if (!spiconfig[spiBusIndex].deviceConfig[deviceIndex].ManualChipSelect)
+            CPU_GPIO_ReservePin(spiconfig[spiBusIndex].deviceConfig->DeviceChipSelect, false);
 
         // Last device on bus then close bus and also remove bus pin reserves
         if (spiconfig[spiDeviceConfig.Spi_Bus].devicesInUse <= 0)
@@ -327,7 +339,7 @@ HRESULT nanoSPI_OpenDeviceEx(
         return hr;
     }
 
-    if (deviceHandle != 0)
+    if (deviceHandle != 0 && !spiDeviceConfig.ManualChipSelect)
     {
         // Reserve chip select pin
         if (CPU_GPIO_ReservePin((GPIO_PIN)spiDeviceConfig.DeviceChipSelect, true) == false)
@@ -382,7 +394,10 @@ HRESULT nanoSPI_CloseDevice(uint32_t handle)
     spiconfig[spiBus].devicesInUse--;
 
     // Unreserve CS pin
-    CPU_GPIO_ReservePin(spiconfig[spiBus].deviceCongfig->DeviceChipSelect, false);
+    if (!spiconfig[spiBus].deviceConfig->ManualChipSelect)
+    {
+        CPU_GPIO_ReservePin(spiconfig[spiBus].deviceConfig->DeviceChipSelect, false);
+    }
 
     // Last device on bus then close bus and also remove bus pin reserves
     if (spiconfig[spiBus].devicesInUse <= 0)
@@ -440,7 +455,7 @@ HRESULT nanoSPI_Write_Read(
 
     return CPU_SPI_nWrite_nRead(
         spiconfig[spiBus].deviceHandles[deviceIndex],
-        spiconfig[spiBus].deviceCongfig[deviceIndex],
+        spiconfig[spiBus].deviceConfig[deviceIndex],
         swrs,
         writePtr,
         wsize,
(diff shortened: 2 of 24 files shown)


src/HAL/Include/nanoHAL_Spi.h
```diff
@@ -18,7 +18,7 @@ struct nanoSPI_BusConfig
 {
     bool spiBusInited;
     int8_t devicesInUse;
-    SPI_DEVICE_CONFIGURATION deviceCongfig[MAX_SPI_DEVICES];
+    SPI_DEVICE_CONFIGURATION deviceConfig[MAX_SPI_DEVICES];
```
networkfusion, line 21: Nice spelling catch!

src/System.Device.Spi/nanoHAL_Spi.cpp
```diff
@@ -112,7 +112,7 @@ static int FindFreeDeviceSlotSpi(int spiBus, GPIO_PIN cs)
         if (spiconfig[spiBus].deviceHandles[deviceIndex] == 0)
             return deviceIndex;
         // Check device chip select not already in use
-        if (spiconfig[spiBus].deviceCongfig[deviceIndex].DeviceChipSelect == cs)
```
networkfusion, line 115: Nice spelling catch!

KasperJSdeVries:
> all existing calls that start and stop an SPI transaction have to toggle the CS pin, if the CS line is different than -1.

@josesimoes you mean that every time a call gets made like `SpiSend` we manually toggle the CS line

josesimoes:
> @josesimoes you mean that every time a call gets made like `SpiSend` we manually toggle the CS line

Correct. So we can keep honoring that behavior.

targets/ChibiOS/_nanoCLR/System.Device.Spi/sys_dev_spi_native_target.h
```diff
@@ -27,6 +27,9 @@ struct NF_PAL_SPI
 
     uint8_t *ReadBuffer;
     uint16_t ReadSize;
+
+    // -1 = no active ChipSelect | -2 = manual Active ChipSelect | 0+ Active chip select pan number
+    int32_t ActiveChipSelect;
```
josesimoes, line 32: "Active" sounds a bit redundant here... why not just simply, ChipSelect? Or ChipSelectPin... 😉

josesimoes:
@Ellerbach yes, that's the idea behind this. The name can be misleading. But that's the point of the PR. 
As it is now (on all platforms) we are NOT compatible with IoT on this aspect.

targets/ChibiOS/_nanoCLR/System.Device.Spi/cpu_spi.cpp
```diff
@@ -534,12 +550,32 @@ HRESULT CPU_SPI_nWrite_nRead(
...
+                    sdev.DeviceChipSelect,
+                    sdev.ChipSelectActive ? GpioPinValue_Low : GpioPinValue_High);
+            }
         }
         else
         // Start an Asyncronous SPI transfer
         // perform SPI operation using driver's ASYNC API
         // Completed on calling Spi Callback
         {
+            if (!sdev.ManualChipSelect)
```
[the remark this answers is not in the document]
KasperJSdeVries: we need to read the value once and then we can stop using it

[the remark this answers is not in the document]
KasperJSdeVries: I wanted to leave out a conversion between uint and int but I guess that is not really necessary as its unlikely any board is going to have more than 2147483648 pins anyway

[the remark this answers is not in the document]
josesimoes: I guess this is an assumption that's pretty safe to take. I don't think that board will be available in a foreseeable future!! 🤣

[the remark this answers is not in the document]
KasperJSdeVries: This change did force me to change some types that were unsigned to their signed variant

src/System.Device.Spi/nanoHAL_Spi.cpp
```diff
@@ -266,20 +266,30 @@ HRESULT nanoSPI_OpenDeviceEx(
     }
 
     // Find if device slot is available and check
-    deviceIndex = FindFreeDeviceSlotSpi(spiBusIndex, spiDeviceConfig.DeviceChipSelect);
+    if (spiDeviceConfig.DeviceChipSelect <= 0)
+    {
+        if (spiconfig[spiBusIndex].devicesInUse > 0)
+            return CLR_E_NOT_SUPPORTED;
```
Ellerbach, line 272: please use explicit opening and closing brackets { }. I know it's not there *yet* everywhere but we're trying to :-)

src/System.Device.Spi/nanoHAL_Spi.cpp
```diff
@@ -266,20 +266,30 @@ HRESULT nanoSPI_OpenDeviceEx(
...
-        }
-        else
+        deviceIndex = FindFreeDeviceSlotSpi(spiBusIndex, spiDeviceConfig.DeviceChipSelect);
+
+        if (deviceIndex < 0)
         {
-            // Return NOT_SUPPORTED when Device already in use. Not really any other relevant exception that's
-            // currently raised in managed code
-            return CLR_E_NOT_SUPPORTED;
+            if (deviceIndex == -1)
```
Ellerbach, line 282: do I understand correctly that if I want to use 2 different SpiDevice with both the chipselect set to -1, it won't work?

src/System.Device.Spi/nanoHAL_Spi.cpp
```diff
@@ -292,7 +302,8 @@ HRESULT nanoSPI_OpenDeviceEx(
         spiconfig[spiBusIndex].devicesInUse--;
 
         // Unreserve CS pin
-        CPU_GPIO_ReservePin(spiconfig[spiBusIndex].deviceCongfig->DeviceChipSelect, false);
+        if (spiconfig[spiBusIndex].deviceConfig[deviceIndex].DeviceChipSelect >= 0)
+            CPU_GPIO_ReservePin(spiconfig[spiBusIndex].deviceConfig->DeviceChipSelect, false);
```
Ellerbach, line 306: explicit brackets please { }

Ellerbach (commented):
Thanks for the improvements! Few questions and an additional one: does the ESP32 implementation need to be adjusted as well?

KasperJSdeVries:
I would assume so. I don't exactly know what would need to be done as haven't worked with an esp32 before.

src/System.Device.Spi/nanoHAL_Spi.cpp
KasperJSdeVries: I don't think that would be worth the trouble. I think in general the user will either manage all of the CS pins for a bus or none.

src/System.Device.Spi/nanoHAL_Spi.cpp
KasperJSdeVries: that's correct, but there is no reason to instantiate 2 SpiDevices if you are managing the CS pins yourself. I could change it so it doesn't throw an exception and just returns the first SpiDevice on the bus.

src/System.Device.Spi/nanoHAL_Spi.cpp
josesimoes: This can be handled per device, no worries.
For STM32 it's a matter of a general configuration to disable CS being handled by the driver (already done here).
For ESP32, the IDF allows up to 3 different devices (with 3 different CS pins) so we're good with enabling this globally and manage it per device.

src/System.Device.Spi/nanoHAL_Spi.cpp
josesimoes: That would be OK looking at this from the native code. Now from C# that wouldn't be correct. If we have 2 different devices we have to have 2 `SpiDevice`  objects.

This requires that the `FindFreeDeviceSlotSpi` needs to be changed so it won't use CS to search for the devices in the collection.

josesimoes:
> Thanks for the improvements! Few questions and an additional one: does the ESP32 implementation need to be adjusted as well?

Absolutely! Any change at this level has to be implemented for all existing platforms before this gets merged.

src/System.Device.Spi/nanoHAL_Spi.cpp
KasperJSdeVries: > This requires that the `FindFreeDeviceSlotSpi` needs to be changed so it won't use CS to search for the devices in the collection.

This has been changed. If I'm correct this should now return a working device regardless. It also allows manually managed and automatically managed devices on the same bus

KasperJSdeVries:
### Progress per target group:
- [x] STM
- [ ] ESP32
- [ ] NXP
- [ ] TI
- [ ] win32

targets/ChibiOS/_nanoCLR/System.Device.Spi/sys_dev_spi_native_target.h
```diff
@@ -27,6 +27,9 @@ struct NF_PAL_SPI
 
     uint8_t *ReadBuffer;
     uint16_t ReadSize;
+
+    // -1 = no active ChipSelect | -2 = manual Active ChipSelect | 0+ Active chip select pan number
+    int32_t ChipSelect;
```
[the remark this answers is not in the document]
KasperJSdeVries: I did this to distinguish between the two, I made this distinction because it would make it easier to activate/deactivate the CS line (calls to `SetPinState`). Looking back at it I don't think it is necessary, but I'm not sure.

josesimoes:
@KasperJSdeVries I've just push the changes we've been working here.
Basically simplified several bits in the platforms using the ChibiOS HAL.
Tested on a couple of M5Stack devices and STM32.

Kindly review so we can have this one merged.

KasperJSdeVries:
Looks good GLTM!

PR #2341, opened 2022-05-25
Optimized opaque Bitmap rectangle fill. BitBlt now accepts both full bitmap and screen coordinates

mikeoliphant:
## Description
This small change optimizes filling of rectangles in Bitmap.DrawRectangle() when the fill color is opaque. It is a modest (about 15%) speed improvement, but it applies to a common use case.

## Motivation
- Following nanoframework/nanoFramework.Graphics#106.

## How Has This Been Tested?
Testing by building and running on the M5Core2 target.

## Types of changes

## Checklist


josesimoes:
15% it's a very nice improvement!

mikeoliphant:
I made another fix which I was intending to submit separately, but I forgot it would get tacked onto this active PR.

The fix is for a bug preventing several of the display drivers from obeying the window size passed to DisplayDriver::BitBlt - only full screen blits were working properly.

The fixes are for ILI9341_240x320, ILI9342_320x240, ST7735S, and ST7789V_240x320.

I have only tested ILI9342_320x240 as that is the only one I have - the code is very similar for the other displays, though, so I'm pretty sure it should be correct.

Without this fix, it is impossible to blit sub-regions of the screen - doing a partial flush of a bitmap, for example.

mikeoliphant:
It just occurred to me that the blit operation is ambiguous between blitting the sub-region (x, y, width, height) of the data to screen position (x, y) and blitting the sub-region (0, 0, width, height) of the data to screen position (x, y).

The existing code was doing the latter. My change makes it do the former. Both operations are useful. Ideally, the BitBlt operation would allow specifying both bitmap source xy and screen destination xy.

mikeoliphant:
From the git history, it looks like the BitBlt interface was originally operating the way I've now changed it back to. Then it was changed at some point - probably to allow for blitting a non-fullscreen bitmap to a different position on the screen? @AdrianSoundy @Ellerbach

josesimoes:
> From the git history, it looks like the BitBlt interface was originally operating the way I've now changed it back to. Then it was changed at some point - probably to allow for blitting a non-fullscreen bitmap to a different position on the screen? 

That is correct. IIRC @Ellerbach  was the one making that improvement. He can comment further.

josesimoes:
> It just occurred to me that the blit operation is ambiguous between blitting the sub-region (x, y, width, height) of the data to screen position (x, y) and blitting the sub-region (0, 0, width, height) of the data to screen position (x, y).

Agreed that both should exist as they fit different usage needs, and both are perfectly valid.

Ellerbach (approved):
Thanks, looks good to me. There are quite few things to optimize and add to this library! Still missing proper rotation and orientation as well as reverting colors for the full screen! So keep coming the improvements!
Last huge one was about adding all those drivers as well as ability to initialize them from the managed side and write directly on the screen for small ram devices as well as writing with dedicated font for the same purpose!
So thanks an keep the improvements coming :-)

mikeoliphant:
I'd be happy to have this PR go through as it addresses my use case (which I think the most common one) - using a bitmap as a full-screen frame buffer.

It does have limitations, though, as I mentioned above:

1) The bitmap *must* be full screen width. BitBlt doesn't take a stride, so the width of the screen is assumed.
2) No screen destination coordinates can be specified independently.

To handle all cases, the BitBlt interface would need to be changed to something like:

BitBlt(int src_x, int src_y, int width, int height, int stride, int dest_x, int_dest_y, CLR_UINT32 data[]) 

This would blit the source rectangle "src_x, src_y, width, height" to position "dest_x, dest_y" on the screen where the actual full width of the bitmap data is "stride".

If you guys think it makes sense, I'd be happy to add that change to this PR. It has more far-ranging consequences, though, so maybe it would be better to consider separately?

josesimoes:
@mikeoliphant thanks for volunteering to the task. 👍🏻 

If I understood all this correctly your change would break a recent change from @Ellerbach  which was precisely to allow making partial updates of a screen so devices low on RAM can deal with displays. To keep supporting that, it would require that extra work that you're kindly offering to tackle.

My take on this would be to have that 2nd part here as it would be the only way forward **without** introducing any breaking changes nor take back features.

@Ellerbach toughts?

mikeoliphant:
Yes - the changes would definitely need to be exposed at the managed level. Fortunately, there aren't too many points of contact - really just the Bitmap Flush() method.

All in all, I think I would only have to change DisplayDriver::BitBlt, GraphicsDriver::Screen_Flush and the Bitmap Flush() managed interface.

josesimoes (approved):
LGTM!

Ellerbach (approved):
Thanks, looks good to me now. So the managed library needs to be updated as well.

mikeoliphant:
I just noticed that this may have caused some issues for the DisplayControl class. I'm looking into it now.

PR #2372, opened 2022-06-15
Fix memory alloc/free for non XIP storage

josesimoes:
## Description
- Add call to free memory on CLR_RT_Assembly::DestroyInstance (there was a memory leak with allocation in ContiguousBlockAssemblies).
- Add call to free memory on failure during ContiguousBlockAssemblies.
- Remove unnecessary checks for pointer not NULL.

## Motivation and Context
- Fixes memory leak when running targets with non XIP storage.

## Types of changes

## Checklist


josesimoes:
@dougboling care to review please?

dougboling:
Jose,
You know the code vastly more than I. However, I worry that DestroyInstance is called in situations where m_header is not a malloc'ed buffer. 

For example, line 395 of src\CLR\CorLib\corlib_native_System_Reflection_Assembly.cpp. In there, m_header seems to be a pointer to a stack object. There, DestroyInstance is called If the routine fails at any point.

Or in LoadDownloadedAssemblies Line 228 in src\CLR\Core\Execution.cpp. There header is returned from GetFirstElement. Is that a malloc'ed buffer? I don't know the code enough to trace the permutations. 

Note, I use header and m_header interchangeably because header is assigned to m_header in CreateInstance.

Is there a need to check some flag such as the XIP state before assuming that m_header points to a native heap object?
Doug

josesimoes:
We may have to add a new flag to the assembly  struct to signal if this should be freed at DestroyInstance or not...

At that point, no mater where it's called from, there is no way to know if it was malloced (or not) or if the storage from where it was loaded from is XIP or not.

josesimoes:
@dougboling I'm back to this PR so we can close it. Those cases you've mentioned (and all the rest) do not require a call to free, as the only execution path that mallocs memory it's the one that you've spotted when loading from non XIP storage. 

The other execution significantly different it's the WIN32 build. Over there the files are loaded in RAM and are freed by the file API.

Please review this when you have the opportunity.

dougboling:
One problem with the change is a new warning compiling Line 1869 of TypeSystem.cpp:
    if (this->m_flags & CLR_RT_Assembly::FreeOnDestroy != 0)

I fixed this by adding an additional set of parens to explicitly define the statement:
    if ((this->m_flags & CLR_RT_Assembly::FreeOnDestroy) != 0)

After that fix, the code compiles clean and appears to work on my system.

Doug

dougboling (changes requested):
Line 1869 of TypeSystem.cpp needs an additional set of parens to encapsulate the AND of the flag before the test for non-zero. As in:
if ((this->m_flags & CLR_RT_Assembly::FreeOnDestroy) != 0)
Without this change, building the file produces a warning.

src/CLR/Core/TypeSystem.cpp
```diff
@@ -1865,6 +1865,12 @@ void CLR_RT_Assembly::DestroyInstance()
     }
 #endif
 
+    // check if header has to be freed (in case the assembly lives in RAM)
+    if (this->m_flags & CLR_RT_Assembly::FreeOnDestroy != 0)
```
josesimoes, line 1869: per @dougboling suggestion

```suggestion
    if ((this->m_flags & CLR_RT_Assembly::FreeOnDestroy) != 0)
```

josesimoes:
@dougboling I've totally missed that warning locally, but the pipeline it's complaining about it too! 😯 
Fixed. Thanks!

PR #2400, opened 2022-07-21
Add GetDevicePublicKey

aromaa:
## Motivation and Context
- Following nanoframework/System.Net#265

## Types of changes

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Added a method to retrieve the device's public key in raw byte array format.
	- Expanded SSL certificate handling capabilities with new functions for public key extraction.

- **Version Update**
	- Updated System.Net assembly version to support new functionality.

- **Technical Improvements**
	- Introduced a new X.509 certificate data structure for handling raw data.
	- Enhanced infrastructure for public key extraction across different platforms.
	- Organized enum types into more appropriate header files for improved maintainability.

### DIFF
--- a/src/DeviceInterfaces/System.Net/sys_net_native_System_Net_Security_CertificateManager.cpp
+++ b/src/DeviceInterfaces/System.Net/sys_net_native_System_Net_Security_CertificateManager.cpp
@@ -81,3 +81,36 @@ HRESULT Library_sys_net_native_System_Net_Security_CertificateManager::
 
     NANOCLR_CLEANUP_END();
 }
+
+HRESULT Library_sys_net_native_System_Net_Security_CertificateManager::GetDevicePublicKeyRaw___STATIC__SZARRAY_U1(
+    CLR_RT_StackFrame &stack)
+{
+    NATIVE_PROFILE_CLR_NETWORK();
+    NANOCLR_HEADER();
+
+    CLR_RT_HeapBlock &ret = stack.PushValueAndClear();
+    HAL_Configuration_X509DeviceCertificate *deviceCert = ConfigurationManager_GetDeviceCertificate();
+
+    if (deviceCert)
+    {
+        X509RawData rawData;
+
+        if (SSL_GetPublicKeyRaw((const char *)deviceCert->Certificate, deviceCert->CertificateSize, &rawData))
+        {
+            CLR_RT_HeapBlock_Array *array;
+
+            NANOCLR_CHECK_HRESULT(
+                CLR_RT_HeapBlock_Array::CreateInstance(ret, rawData.len, g_CLR_RT_WellKnownTypes.m_UInt8));
+
+            array = ret.DereferenceArray();
+
+            memcpy(array->GetFirstElement(), rawData.p, rawData.len);
+
+            platform_free(rawData.p);
+        }
+
+        platform_free(deviceCert);
+    }
+
+    NANOCLR_NOCLEANUP();
+}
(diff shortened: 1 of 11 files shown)


src/DeviceInterfaces/System.Net/sys_net_native_System_Net_Security_CertificateManager.cpp
```diff
@@ -81,3 +81,31 @@ HRESULT Library_sys_net_native_System_Net_Security_CertificateManager::
...
+    if (deviceCert)
+    {
+        CLR_RT_HeapBlock_Array *array;
+
+        NANOCLR_CHECK_HRESULT(
+            CLR_RT_HeapBlock_Array::CreateInstance(ret, deviceCert->CertificateSize, g_CLR_RT_WellKnownTypes.m_UInt8));
+
+        array = ret.DereferenceArray();
+
+        memcpy(array->GetFirstElement(), deviceCert->Certificate, deviceCert->CertificateSize);
```
Ellerbach, line 103: here you only return the first element. Why not returning the array? You may have multiple certificates uploade, so you may want to get all the associated names.
I understand that later it can be a pain to know which one is associated with which certificate. So maybe on top of that, it should return as well the certificate name or equivalent?

aromaa: Right. I decided to use the identical method that was being called within the SSL handshake to determine the certificate. Returning the array of certificates here would require the caller to know which one is actually being used then.. Hmm.. I guess more correct approach would be to return the used certificate from SslStream?

aromaa: Also that approach would require changes to for example the Azure IoT libraries to expose the underlying cert used as you have no access to SslStream.

josesimoes: Couple of comments on this:
1. The internal storage as only room for one device certificate.
2. We can not have it implemented this way. It would be a serious security flaw! Device certs can not be exposed under any circumstances. That's why on everything that's here today it never "leaves" the native layer and it's accessed directly by the network stack. Nothing else. 

OK with exposing the public key, which is all this is about. But that's it.

I'll give it some though and provide more feedback during next week because right now I'm overloaded with other tasks.

aromaa: > We can not have it implemented this way. It would be a serious security flaw!

Yeah, was looking at this today again and realized that I forgot that the public cert & private key are stored under the same variable. I'm gonna take a look at that we strip off the private key.

josesimoes: If what we need is the public key (and we are in agreement that's a nice improvement) from the device cert. That exactly what we'll be exposing. Nothing more, nothing less. 😉

aromaa: Yes. That's my intention and that's what the method is named after all. The private key is indeed inaccessible due to security reasons and I don't intend (or even want!) to expose that.

aromaa:
Okay, I somehow did something wrong while rebasing this to fix the force push conflict and messed up the whole commit history and lost my own while at it. *sigh* Anyways, fixed it.

networkfusion:
@aromaa Is this PR still valid given all the changes?

aromaa:
This needs a bit updating to stub the new platforms and a bit of updating. I'm planning to do that in the future as I will be needing to update my devices soon for the new Azure certificate so the branch will be up to date after that.

aromaa:
Actually there wasn't much else to do as the new platforms are already falling back to the existing stubs. Fixed few copy & paste mistakes. Still seems to work just fine on my ESP32.

aromaa:
Rebased and updated the System.Net native bindings. Everything seems to still work just fine on my ESP32.

PR #2425, opened 2022-08-30
Added support for LilygoTWatch2021 (ESP32-PICO-V3-02 + GC9A01) 

amzak:
## Description

This PR adds support for LilygoTWatch2021.
The device is a ESP32-PICO-V3-02 with 2Mb embedded PSRAM and a touch display with GC9A01 controller.

- added sdk config file sdkconfig.default_pico_rev3_psram; (copied from a pico config, modified psram related options)
- added preset LilygoTWatch2021_preset; (copied from LilygoTWatch2020)
- GC9A01 native graphics driver implementation;
- stub for CST816S touch controller;

## Motivation and Context
Was not supported before.

## How Has This Been Tested?
Locally on actual device - screen works.

## Types of changes

## Checklist


src/nanoFramework.Graphics/Graphics/Displays/GC9A01_240x240_SPI.cpp
```diff
@@ -0,0 +1,271 @@
+#include "Graphics.h"
```
Ellerbach, line 1: please add the standard header in this file.

Ellerbach: Additional question: what is the difference between this driver and the other existing ones? Any pointer on the reference or differences with the other ones?

src/nanoFramework.Graphics/Graphics/Displays/GC9A01_240x240_SPI.cpp
```diff
@@ -0,0 +1,271 @@
...
+    Attributes.BitsPerPixel = 16;
+    g_DisplayInterface.GetTransferBuffer(Attributes.TransferBuffer, Attributes.TransferBufferSize);
+    return;
+}
+
+bool DisplayDriver::Initialize()
+{
+    SetupDisplayAttributes();
+
+    g_DisplayInterface.SendCommand(1, GC9A01_CMD::InnerRegisterEnable2);
```
Ellerbach, line 92: would be nice to have a link on from where all those magic commands are coming from

src/nanoFramework.Graphics/TouchPanel/Devices/CST816S.cpp
```diff
@@ -0,0 +1,37 @@
+#include "TouchDevice.h"
```
Ellerbach, line 1: please add as well the standard header here

src/nanoFramework.Graphics/TouchPanel/Devices/CST816S.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "TouchInterface.h"
```
Ellerbach, line 2: So seems like a placeholder. And for touch, you have quite a few implemented in C# pure managed code. See https://github.com/nanoframework/nanoFramework.IoT.Device/tree/develop/devices#touch-sensors. It may be easier in C# as so far those are just prototypes in native.

Ellerbach (commented):
Nice addition! Few comments.

Ellerbach:
@josesimoes the build is failing but not because of the PR. Seems that one of the device is missing some recent changes on the GCC version.

josesimoes:
> @josesimoes the build is failing but not because of the PR. Seems that one of the device is missing some recent changes on the GCC version.

It's a matter of re-running the job. This started to happen last week and only for this target, which is using the resources and scripts as all the other STM ones. To make things worst this is always failing on the unpacking of the compressed file with the toolchain.

src/nanoFramework.Graphics/Graphics/Displays/GC9A01_240x240_SPI.cpp
amzak: Well, I guess it's very similar to ILI9341 or ST7735S, except for the initialization sequence, resolution, and may be some command's mappings, but the core commands look the same, like
MemoryWrite, Column/PageAddressSet, power control etc.

src/nanoFramework.Graphics/Graphics/Displays/GC9A01_240x240_SPI.cpp
amzak: That is not an easy thing to do) The only datasheet I could find is a preliminary version and it is not complete, for instance, the initialization sequence is not documented at all.
I've reviewed some code samples and alternative driver libraries for this controller, so PR's implemetation was based on this.
Should I attach the datasheet somewhere? It looks like this [GC9A01A.pdf at waveshare](https://www.waveshare.com/w/upload/5/5e/GC9A01A.pdf,) and manufacturer's site doesn't help.

src/nanoFramework.Graphics/TouchPanel/Devices/CST816S.cpp
amzak: My first attempt was not to include touch at all, but then the graphics library does not compile. May be there is a way to get around that?
I guess, I could have used XPT2046, which is a placeholder too, but it would be misleading.
Regarding managed implementation, looks like there is no implementation for CST816S and I have some concerns about speed and power consumption, but as a start I could implement it.

Ellerbach: That's fine, it's a placeholder anyway. Implementing the CST816S in manage code will be super straight forward. Seems very easy and you can reuse the existing ones. From existing drivers, speed is largely ok, consumption wise, it's like if it would be in native, it would fetch all the time the I2C bus to get the data from the driver. So not much less, not much more I would say.

src/nanoFramework.Graphics/Graphics/Displays/GC9A01_240x240_SPI.cpp
Ellerbach: All those are so similar and different at the same time, it's a bit frustrating as it does somehow prevent to easily create some universal drivers and some auto detection. But thanks again for adding this one!

josesimoes (commented):
My review comments are focused on the sdk config file.

From the looks of it (and correct my if I'm wrong) the relevant changes from the existing `sdkconfig.default_pico` are enabling support for PSRAM and enabling BLE.

All the other changes seem to be unintentional from your local testing.
Even the fact that you named the config `rev3` is not correct because that is not reflected on the config option.

If the above is correct, my suggestion goes towards:
1. Changing the existing config to enable support for PSRAM (as that won't hurt devices without it)
2. Enable BLE for the new configuration and rename it to sdkconfig.default_pico_ble.

amzak:
@josesimoes Yes, you are absolutely right. The most of these changes should not have got into the PR. My bad
As you've suggested, I enabled psram in default pico config and made a copy of it with bluetooth enabled.

josesimoes:
@amzak I've just add some missing config blocks for PSRAM and Nimble and updated the branch.
Should be OK for merge now.

josesimoes (approved):
LGTM!

josesimoes:
@amzak care to share the URL for the 2021 version?   I was adding this to the targets list but couldn't find it on lilygo website...

josesimoes:
@amzak one more thing: I've just added this to be build in Comunity Repo nightly build. Thing is that it's failing...
https://dev.azure.com/nanoframework/Community-Targets/_build/results?buildId=36294&view=logs&j=3e2e46fb-05e9-57f5-652a-47ac5d5f5c51&t=70a0fb5d-722f-5a09-dfd3-6a4cc5fd6cac

Any chance you can take a look and find out what's happening please?

amzak:
@josesimoes, sure, I'll look into it
Regarding the link, there is this [product page](http://www.lilygo.cn/prod_view.aspx?TypeId=50053&Id=1377&FId=t3:50053:3) and [github repo](https://github.com/Xinyuan-LilyGO/T-Watch-2021)

PR #2438, opened 2022-09-26
Add I2S for ESP32

BWalti:
## Description
- Adjust CMake for ESP32 and I2S.
- Fixed some implementation details (device pin mapping).
- Some adaptations in "init" according to the ESP IDF 4.4.2 documentation.
- Update I2S assembly declaration.

## Motivation and Context

- Following nanoframework/System.Device.I2s#3.
- Addresses nanoframework/Home#571.

## How Has This Been Tested?

- With the AudioPlayer test application from System.Device.I2s -> "add-basic-test-app" branch.
--> I can successfully play an audio file and it is quite good. 
- I tested Read operation with the I2sApplication in the same repo - but without using a microphone as I have none. At least I did not get any errors and some data has been produced / read (array was not zeros). But no guarantee that this will work in every case.

## Types of changes

## Checklist


CMakePresets.json
```diff
@@ -119,6 +121,7 @@
                 "API_System.Device.Gpio": "ON",
                 "API_System.Device.Spi": "ON",
                 "API_System.Device.I2c": "ON",
+                "API_System.Device.I2s": "OFF",
```
BWalti, line 124: maybe I could leave all the "default OFFs" away?

targets/ESP32/_include/Esp32_DeviceMapping.h
```diff
@@ -45,13 +45,14 @@ enum Esp32_MapDeviceType
     DEV_TYPE_SERIAL,
     DEV_TYPE_LED_PWM,
     DEV_TYPE_ADC,
-    DEV_TYPE_DAC,
     DEV_TYPE_I2S,
+    DEV_TYPE_DAC,
```
BWalti, line 49: this order change is required in my understanding so that the `DeviceFunction` enum values for I2S are matching

targets/ESP32/_nanoCLR/System.Device.I2s/sys_dev_i2s_native_System_Device_I2s_I2sDevice.cpp
```diff
@@ -148,51 +132,113 @@ HRESULT Library_sys_dev_I2s_native_System_Device_I2s_I2sDevice::Write___VOID__SZ
...
+        }
+
+        // Set the Bus parameters
+        NANOCLR_CHECK_HRESULT(SetI2sConfig(bus, pConfig));
+
+        // dereference the write and read SpanByte from the arguments
+        writeSpanByte = stack.Arg1().Dereference();
+        if (writeSpanByte != NULL)
+        {
+            writeData = writeSpanByte[SpanByte::FIELD___array].DereferenceArray();
```
BWalti, line 174: now here starts all the wild c++ things I don't understand in depth - or I am not sure results in what we actually want to achieve :)

Ellerbach: Here you have a good explanation (if not good, let me know, I wrote it :-D): https://docs.nanoframework.net/content/architecture/nanoclr-stub-args.html#dereferencing-an-object

In short, the writeSpanByte is an object. You first need a point on it (that's le line 171). Then you need access to the array in it, that's the purpose of the line 174. 

First the `writeSpanByte[SpanByte::FIELD___array]` gives a pointer to the field array and then doing `.DereferenceArray()` gives you the actual array.

If you would use a real array rather than a SpanByte, it would be a bit difference. But in our case, we're using the SpanByte, so getting access to the array and then pointing on the right element is what needs to be done.

If not clear enough, let me know!

CMakePresets.json
Ellerbach: You can make it ON for the devices that have actually the support for it.

.vscode/cmake-variants.TEMPLATE.json
```diff
@@ -109,6 +109,7 @@
           "API_System.Device.Dac": "OFF-default-ON-to-add-this-API",
           "API_System.Device.Gpio": "OFF-default-ON-to-add-this-API",
           "API_System.Device.I2c": "OFF-default-ON-to-add-this-API",
+          "API_System.Device.I2s": "OFF-default-ON-to-add-this-API",
```
josesimoes, line 112: We're not using the cmake-variants anymore. Now it's all CMake templates. 
Nevertheless, let this change here, so when this is ported over to the docs website it will go along.

CMake/Modules/FindSystem.Device.I2s.cmake
```diff
@@ -8,20 +8,21 @@ set(BASE_PATH_FOR_THIS_MODULE ${BASE_PATH_FOR_CLASS_LIBRARIES_MODULES}/System.De
 
 
 # set include directories
-list(APPEND System.Device.I2s_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/src/CLR/Core)
-list(APPEND System.Device.I2s_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/src/CLR/Include)
-list(APPEND System.Device.I2s_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/src/HAL/Include)
-list(APPEND System.Device.I2s_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/src/PAL/Include)
+list(APPEND System.Device.I2s_INCLUDE_DIRS ${PROJECT_SOURCE_DIR}/src/CLR/Core)
```
josesimoes, line 11: All these have to be `CMAKE_SOURCE_DIR` ...

targets/AzureRTOS/MicroBit/MICRO_BIT_2/cmake-variants.json
```diff
@@ -49,6 +49,7 @@
           "API_System.Math": "OFF",
           "API_System.Device.Gpio": "ON",
           "API_System.Device.I2c": "OFF",
+          "API_System.Device.I2s": "OFF",
```
josesimoes, line 52: No need to touch this one, I'm guessing... no work being done for microbit, right?

targets/FreeRTOS/NXP/NXP_MIMXRT1060_EVK/README.md
```diff
@@ -50,6 +50,7 @@ Pre-configured cmake-variants.json for this board
           "API_System.Device.Gpio": "ON",
           "API_System.Device.Spi": "OFF",
           "API_System.Device.I2c": "OFF",
+          "API_System.Device.I2s": "OFF",
```
josesimoes, line 53: Same here. No need to touch this.

targets/ESP32/_nanoCLR/System.Device.I2s/sys_dev_i2s_native_System_Device_I2s_I2sDevice.cpp
josesimoes: Looks good to me as you have it! If it's not outputting then something else is missing.

CMakePresets.json
BWalti: but then I'd also have to implement the native binding, right?
I fear that is going to take me a long time - maybe I focus on one platform for now and check back later to implement the same feature on other platforms maybe?

CMake/Modules/FindSystem.Device.I2s.cmake
BWalti: where exactly is the difference? and why is the file generated "wrongly" therefore?
I assume the project source dir points to the "checked out" root of the nf-framework repository? and the cmake too?

targets/ESP32/_nanoCLR/System.Device.I2s/sys_dev_i2s_native_System_Device_I2s_I2sDevice.cpp
BWalti: could the wrong "FindSystem.Device.I2s.cmake" usage of `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR` explain the (probably) wrong behavior?

BWalti: @Ellerbach : I am just about to read through the linked documentation and for starters I'd wish a [link](https://github.com/nanoframework/CoreLibrary/blob/main/nanoFramework.CoreLibrary/System/SpanByte.cs/) to the SpanByte (even though I found it pretty easily within the repository - but others might be tempted to check the official dotnet repo implementation - if there is even such a class).

and next: in my understanding `SpanByte` is therefore a "wrapper" around `byte[]` which includes pointer to a part of the array. as long as I use the "whole" `SpanByte` backing buffer array I assume performance wise this is not going to be faster in any way?

Would therefore the idea be to allocate a proper "SpanByte buffer" and then write data to sub-segments and hand over the SpanByte to the native implementation so that we won't have to GC the byte[] all the time otherwise?

BWalti: btw. I saw this morning that I Dereferenced the same backing array twice - removed the duplicated code and simplified it a bit. I am going to check now, if all these minor changes (CMake, code cleanups, etc) helped in any way :)

BWalti: uh I got noise out of the speaker 🥳. now up to getting "real audio" to check if it works 😊

CMake/Modules/FindSystem.Device.I2s.cmake
josesimoes: This has now being resolved and the correct variable is being used. The CMake template being generated in the C# project it's correct. For some reason you've changed the CMake variable names in the original commit.

targets/ESP32/_nanoCLR/System.Device.I2s/sys_dev_i2s_native_System_Device_I2s_I2sDevice.cpp
josesimoes: Happy to know that you're making progress!! 👍🏻

josesimoes: On the SpanByte, that class doesn't exist on .NET, only in nanoFramework. It was added as a "replacement" to the original one because of lack of support for generics. Having it makes the code conversion and reusability much smoother.

If I understand your comment above correctly: that's the way to go. Create a SpanByte, fill in with whatever data you want to send to I2S and and it over to the native code.
If you need inspiration check the code for I2C or SPI.

BWalti:
Audio Output works - at least with a wav file (contained in the [PR](https://github.com/nanoframework/System.Device.I2s/pull/3)) with 16 bits, 16kHz and Mono. I have no Microphone, therefore no implementation done in that part.
I also was heavily inspired by [microPythons implementation](https://github.com/micropython/micropython/blob/master/ports/esp32/machine_i2s.c). As that is MIT licensed, I hope this is ok. I most probably will have to adapt the C code for formatting issues.. I assume the bot will help me out there :)

josesimoes:
@BWalti check the doc here on how to have the machine working for you on formatting 😉 
https://docs.nanoframework.net/content/contributing/cxx-coding-style.html

josesimoes:
@BWalti merge conflicts are mostly because of case changes in the assembly names. You can fix those very easily by choosing the correct versions.

BWalti:
> @BWalti check the doc here on how to have the machine working for you on formatting 😉 https://docs.nanoframework.net/content/contributing/cxx-coding-style.html

interesting: I was using the devcontainer and assumed everything "I need" is inside. I guess I'll create another PR afterwards to extend the ESP32 debian dockerfile (to `apt install clang-format`) and also include the clang extension within vs code when using devcontainer. At least that's what I did locally, and formatting is now working.
Maybe I can also extend the other dockerfiles accordingly. I assume all of them are based on debian (where installing `clang-format` is as simple as `apt install clang-format`), but have to check.

josesimoes:
> interesting: I was using the devcontainer and assumed everything "I need" is inside. I guess I'll create another PR afterwards to extend the ESP32 debian dockerfile (to `apt install clang-format`) and also include the clang extension within vs code when using devcontainer. 

That would be an excellent improvement! Please go for it!

> At least that's what I did locally, and formatting is now working. Maybe I can also extend the other dockerfiles accordingly. I assume all of them are based on debian (where installing `clang-format` is as simple as `apt install clang-format`), but have to check.

Yes, the "base" image for all dev containers is the same.

CMakePresets.json
BWalti: enabled I2S for ESP32 variants

BWalti:
@josesimoes / @Ellerbach : what is in your opinion still open to be able to merge this and the related managed class library PR? Would be nice if we could merge this while Haktoberfest is still on going, but also ok if not. No pressure 🙂

josesimoes:
@BWalti no problem with that! Please address the issue with the build of FEATHER_S2 before this gets merged.

BWalti:
ah ok, I see. that is actually a compilation issue my changes caused. I thought it was again some kind of pipeline problem. Allright, will look into that asap. thank you!

Ellerbach:
@BWalti great to see all this. Great to see we are so close to merge! Great work! 👏

josesimoes (approved):
LGTM!

PR #2448, opened 2022-10-03
Lilygo2021 target fix

amzak:
## Description

This PR fixes the build error for Lilygo2021 target. 

## Motivation and Context

It turns out, that the current build configuration with graphics and bluetooth api enabled doesn't fit into iram0. 
It can't fit even with all the optimizations enabled, as advised by Espressif.
But it can fit with REV3 sdk config. 
Since the device is in fact REV3 cpu, I've added a new sdk config, and modified the target profile.

Probably, configuration sdkconfig.default_pico_ble should be deleted, since it's not used, but I'm not sure.

## How Has This Been Tested?

Tested locally.

## Types of changes

## Checklist


josesimoes (changes requested):
Looking at this default_pico_ble sdk config was added by yourself a few days ago.
And it's used only in the Lilygo2021 cmake preset.

I understand your point on making this a rev3 config. Makes total sense.
Let's, instead of creating a new sdk config, rename the sdk config from `default_pico_ble` to this new name `default_pico_ble_rev3`.

I've compared the contents of the 3 pico SDKs and found some differences which shouldn't be there (for consistency). Please address those so we can have the sdk configs nicely aligned. 😉

josesimoes (commented):
I've just aligned a few of the options with the other SDK config files.
Apart from that it looks OK.

josesimoes:
@amzak if you can test these out, please do and report back. After doing that and comiting the changes, this is good to merge.

amzak:
@josesimoes, can you elaborate, please. 
I'm a little confused now, where are these changes come from?
For instance, the suggested change:

```
# CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_DEBUG is not set
to
# CONFIG_BOOTLOADER_COMPILER_OPTIMIZATION_DEBUG=y
```

But at least 5 other configs have that same line.
What should I use as the sample of the correctly aligned sdkconfig file?
Should I align pico_ble_rev3 the same way?
Just trying to understand to avoid that in the future.

josesimoes:
@amzak I was comparing with the the `sdkconfig.default_pico` and `sdkconfig.default.esp32`.
The idea is to have the various configs aligned throughout the files to make it easier to compare.

You had several which here moved a few lines down/up or just commented with the different format.

josesimoes (approved):
@amzak I can see you've made the changes even before my explanation. 😅 
Hope it makes sense to you.

It's looks OK now. I'll merge it. Thanks for improving this.

PR #2456, opened 2022-10-31
Increase reserved memory for IDF

AdrianSoundy:
## Description
Reported as a problem with wifi ScanAsync giving an Invalid operation exception. This was caused due to Wifi not running as it failed to initialise due to out of memory error.

Also had a problem compiling locally the ESP32_PICO target as it got into a loop including ffconf.h due to this in esp32\ffconf.h when sdcard not included. So for now just included SD card on pico

esp32\ffconf.h
```
#if (HAL_USE_SDC != TRUE)
// need this include here when not using SDCARD so it can load the one from IDF
#include <ffconf.h>
#endif
```

## Motivation and Context
- Fixes nanoFramework/Home#1166

## How Has This Been Tested?
Tested with Pico and Olimex boards using Wifi scan sample

## Types of changes

- [ ] Improvement (non-breaking change that improves a feature, code or algorithm)
- [x] Bug fix (non-breaking change which fixes an issue with code or algorithm)
- [ ] New feature (non-breaking change which adds functionality to code)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Config and build (change in the configuration and build system, has no impact on code or features)
- [ ] Dev Containers (changes related with Dev Containers, has no impact on code or features)
- [ ] Dependencies (update dependencies and changes associated, has no impact on code or features)
- [ ] Documentation (changes or updates in the documentation, has no impact on code or features)

## Checklist

- [x] My code follows the code style of this project (only if there are changes in source code).
- [ ] My changes require an update to the documentation (there are changes that require the docs website to be updated).
- [ ] I have updated the documentation accordingly (the changes require an update on the docs in this repo).
- [x] I have read the [CONTRIBUTING](https://github.com/nanoframework/.github/blob/main/CONTRIBUTING.md) document.
- [x] I have tested everything locally and all new and existing tests passed (only if there are changes in source code).

PR #2460, opened 2022-11-01
Add build option to reserve internal RAM for IDF allocation

josesimoes:
## Description
- Add CMake build option `ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION`.
- Update CMake and ESP32 target platform accordingly.
- Update memory allocation define.
- Update CmakePresets for all PICO and Ethernet targets to reserve 10k of internal RAM.

## Motivation and Context
- Exposes option to reserve RAM for IDF allocation.
- This has come up in the review of PR #2456 which was fixing nanoframework/Home#1166.

## Types of changes

## Checklist

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -655,7 +655,6 @@
             "hidden": true,
             "cacheVariables": {
                 "SDK_CONFIG_FILE": "sdkconfig.default_nopsram.esp32",
-                "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "10",
                 "NF_BUILD_RTM": "OFF",
                 "NF_FEATURE_DEBUGGER": "ON",
                 "NF_FEATURE_RTC": "ON",
@@ -726,7 +725,6 @@
                 "NF_BUILD_RTM": "OFF",
                 "NF_FEATURE_DEBUGGER": "ON",
                 "TARGET_SERIAL_BAUDRATE": "115200",
-                "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "10",
                 "NF_FEATURE_RTC": "ON",
                 "API_System.IO.FileSystem": "ON",
                 "API_nanoFramework.Device.OneWire": "ON",
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -632,7 +632,6 @@
             "hidden": true,
             "cacheVariables": {
                 "SDK_CONFIG_FILE": "sdkconfig.default_ble_rev3.esp32",
-                "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "10",
                 "NF_BUILD_RTM": "OFF",
                 "NF_FEATURE_DEBUGGER": "ON",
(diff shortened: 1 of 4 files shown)


CMakePresets.json
```diff
@@ -630,6 +632,7 @@
             "hidden": true,
             "cacheVariables": {
                 "SDK_CONFIG_FILE": "sdkconfig.default_ble_rev3.esp32",
+                "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "10",
```
AdrianSoundy, line 635: Rev3 and bluetooth so no required.

CMakePresets.json
```diff
@@ -653,6 +656,7 @@
             "hidden": true,
             "cacheVariables": {
                 "SDK_CONFIG_FILE": "sdkconfig.default_nopsram.esp32",
+                "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "10",
```
AdrianSoundy, line 659: Config without psram so not required

CMakePresets.json
```diff
@@ -721,6 +727,7 @@
                 "NF_BUILD_RTM": "OFF",
                 "NF_FEATURE_DEBUGGER": "ON",
                 "TARGET_SERIAL_BAUDRATE": "115200",
+                "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "10",
```
AdrianSoundy, line 730: This device is a pico D4 without psram but this entry is using a normal esp32 nopsram config which seems wrong. If right then doesn't need the ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION

AdrianSoundy (changes requested):

AdrianSoundy:
The builds are failing because its not finding the default value of ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION (0)
So we get 
```
C:/nanof/nf-interpreter/build/targets/ESP32/ESP32/target_platform.h:13:47: error: 'OFF' was not declared in this scope
[build]  #define ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION OFF
```

CMakePresets.json
josesimoes: You are correct. Let's fix this on another PR.
Regarding this one: as this is a PICO without PSRAM it should have these reserved 10k, correct?

CMakePresets.json
josesimoes: I was under the impression that these extra 10k where required for devices WITHOUT PSRAM, that's why I've added it here...

AdrianSoundy: It's for  Rev 1 devices that have psram configured which uses over 10K of ram for the psram fixes library.
So really for rev 1 devices which have psram configured only. Like the PICO which had the psram enabled causing the problem.

CMakePresets.json
AdrianSoundy: Shouldn't need extra ram as current config without psram.

AdrianSoundy:
Other than changes above looks good to me,

PR #2513, opened 2023-01-15
Fix LockEventMutex referencing a disposed pointer

alberk8:
## Description

- On repeated BLE Scan and connect to a device cycle, the system will randomly crash and a reboot will occur.

## Motivation and Context

- It fixes the problem of the BLE Client crashing randomly while switching between scanning and connecting a Server/Central.

## How Has This Been Tested?

- One  [BLE Server](https://github.com/alberk8/nanoFramework-BleServer) app is loaded on one ESP32 WROOM/WROVER hardware. 
- One or Two [BLE Client](https://github.com/alberk8/nanoFramework-BleClient) app is load on to ESP32 WROVER hardware(s).
- Connect to the hardware via Serial Terminal.
- Tested the setup for more than 2 weeks continuously.

## Types of changes

## Checklist


targets/ESP32/_nanoCLR/nanoFramework.Device.Bluetooth/nimble_utils.cpp
```diff
@@ -209,6 +209,11 @@ HRESULT PushEmptyBufferToStack(CLR_RT_StackFrame &stack)
 
 bool LockEventMutex()
 {
+    if (ble_event_data.mutex == NULL)
```
josesimoes, line 212: Shouldn't this check also happen in `ReleaseEventMutex()` right bellow this one?

josesimoes: Also what are the implications on the `LockEventMutex` callers when receiving false? Is the code prepared to deal with this?

alberk8: I have not encountered any calls to a null or disposed ReleaseEventMutex before but it should not do any harm.   When  LockEventMutex is false the failure if any is handled in the managed code.

AdrianSoundy:
The problem isn't with the LockEventMutex(), it is due to the failure to allocate the semaphore in DeviceBleInit()
```
ble_event_data.mutex = xSemaphoreCreateMutex();
```
Should be:
```
ble_event_data.mutex = xSemaphoreCreateMutex();
if (ble_event_waitgroup == NULL || ble_event_data.mutex == NULL)
{
    return false;
}
```

So that an out of memory exception is raised.
Could also improve the Device_ble_dispose() so it only deletes waitgroup and mutex when not NULL.

alberk8:
> The problem isn't with the LockEventMutex(), it is due to the failure to allocate the semaphore in DeviceBleInit()
> 
> ```
> ble_event_data.mutex = xSemaphoreCreateMutex();
> ```
> 
> Should be:
> 
> ```
> ble_event_data.mutex = xSemaphoreCreateMutex();
> if (ble_event_waitgroup == NULL || ble_event_data.mutex == NULL)
> {
>     return false;
> }
> ```
> 
> So that an out of memory exception is raised. Could also improve the Device_ble_dispose() so it only deletes waitgroup and mutex when not NULL.

The   [vSemaphoreDelete(ble_event_data.mutex)](https://github.com/nanoframework/nf-interpreter/blob/92ae087d2c2710515009cd5d6a6588e341471448/targets/ESP32/_nanoCLR/nanoFramework.Device.Bluetooth/esp32_nimble.cpp#L478) does not null the pointer,  I have not found a way to test if the Mutex is deleted or not.

AdrianSoundy:
Shouldn't need to null ptr first unless the dispose is also improved only delete if not null

alberk8:
> Shouldn't need to null ptr first unless the dispose is also improved only delete if not null

In my PR, I have null the ble_event_data.mutex  after the [vSemaphoreDelete(ble_event_data.mutex)](https://github.com/nanoframework/nf-interpreter/blob/92ae087d2c2710515009cd5d6a6588e341471448/targets/ESP32/_nanoCLR/nanoFramework.Device.Bluetooth/esp32_nimble.cpp#L478) , is that a good approach?

AdrianSoundy:
The code needs to be improved for these low memory situations, so things get cleaned up in  Device_ble_dispose.

So Device_ble_dispose()
```
    if (ble_event_waitgroup)
    {
        vEventGroupDelete(ble_event_waitgroup);
    }
    if (ble_event_data.mutex)
    {
        vSemaphoreDelete(ble_event_data.mutex);
    }

    ble_event_waitgroup = NULL;
    ble_event_data.mutex = NULL;
```

And DeviceBleInit()

```
ble_event_waitgroup = xEventGroupCreate();
ble_event_data.mutex = xSemaphoreCreateMutex();
if (!ble_event_waitgroup || !ble_event_data.mutex)
{
    return false;
}
```

alberk8:
@AdrianSoundy,  I have applied the suggested changes and let it run overnight and it looks good.  No crash or reboot so far.

Ellerbach (approved):
looks good to me

AdrianSoundy:
All looks good to me also. Will Merge

PR #2574, opened 2023-02-17
Adjusting native definition of System.Device.Wifi

Ellerbach:
## Description

Adjusting native definition of System.Device.Wifi

## Motivation and Context

- following the changes made on the managed side on the helper. Helper is part of the code, so it has changes the native signature.

## Types of changes

## Checklist

### DIFF
--- a/src/System.Device.Wifi/sys_dev_wifi_native.cpp
+++ b/src/System.Device.Wifi/sys_dev_wifi_native.cpp
@@ -74,12 +74,14 @@ static const CLR_RT_MethodHandler method_lookup[] =
     NULL,
     NULL,
     NULL,
+    NULL,
+    NULL,
 };
 
 const CLR_RT_NativeAssemblyData g_CLR_AssemblyNative_System_Device_Wifi =
 {
     "System.Device.Wifi",
-    0x1C1D3214,
+    0x00A058C6,
     method_lookup,
     { 100, 0, 6, 4 }
 };


josesimoes (approved):
Yes this was behind! 
Formally we should have the version bumped too, but that would require a new release of the C# library, so let's keep it as it is.

DaveSchmid:
Wifi doesn't work anymore!
ESP32_REV0-1.8.0.1043

Wifi: 1.5.54
(This package requires a target with System.Device.Wifi v100.0.6.4 (checksum 0x00A058C6).)

VS Extension
15:23:37.091 [Version mismatch for System.Device.Wifi. Need v1.5.54.0, checksum 0x00A058C6.]
15:23:37.092 [The connected target has v100.0.6.4, checksum 0x1C1D3214.]

josesimoes:
@DaveSchmid you have to update the fw of the ESP32 you're using.

DaveSchmid:
@josesimoes  By firmware do you mean ESP32_REV0-1.8.0.1043?
This is the newest one you can download. But there is no suitable for the latest Wifi: 1.5.54.

It would be good if there is a table which version of which library fits together. And which checksum?

josesimoes:
> @josesimoes By firmware do you mean ESP32_REV0-1.8.0.1043? This is the newest one you can download. But there is no suitable for the latest Wifi: 1.5.54.


Yes. But I'm looking at the versions in Cloudsmith and it seems there is a mismatch on the fw versions and files... Investigating.

josesimoes:
> It would be good if there is a table which version of which library and firmware fits together. And which checksum?

This has been asked several times in the past. That table it's "impossible" to maintain and manage. Best approach is to always keep up with the latest or maintain (and possibly) freeze to a set of known versions of fw+nugets.

Anyway the package manager in VS always tells you what fw version it requires.

https://docs.nanoframework.net/content/faq/working-with-vs-extension.html#when-deploying-an-application-i-get-a-message-like-couldnt-find-a-valid-native-assembly-required-by-complaining-that-it-cant-find-a-native-assembly-what-can-i-do-to-fix-this

DaveSchmid:
@josesimoes I don't think that's impossible. Most of the information about the library relationships is in NuGet and CloudSmith available.
These dependencies in NuGet work quite well. Why are the others in Cloudsmith? money?
Is it not possible to read this metadata via api and link it to a table?
I think it would be cool if you could select the device target in the device explorer and the appropriate firmware for the selected mscorlib from nuget would be automatically uploaded to the device.

josesimoes:
@DaveSchmid please note the air quotes around _impossible_ 😉 

You have .NET assembly version, that correspond to NuGet versions, and that's what NuGet is able to match.
Then for each NuGet version you have a series of fw versions that match the required assembly version & checksum.
Those are independent. One can bump without the other.

At the firmware end you have a fw version that "fits" a set of native assembly versions.

The complication here resides on the immense list and variation possibilities that you can get. And it gets even worst if you search from firmware to NuGet.

NuGet can only store nugets. We can't possibly distribute firmware packages there.
Cloudsmith is the only Open Source friendly package hosting for raw binary distribution.

You can read some metadata with APIs from both providers but it requires _some_ degree of parsing to get the job done. And even then, it will always give you a list of possible hits. You won't even get a single hit (unless that's the very first version after a bump).

Know that @Feiko (and others) has already given some thought on this.
Suggest that you start a discussion here on GitHub or on Discord about this. This conversation here has the potential to get lost or hard to find as it is hijacking a PR which is completely unrelated.

DaveSchmid:
You're right, this is the wrong place for this discussion. And I think I need to get a better picture of how the non-.NET part works first to really be able to say anything.

PR #2583, opened 2023-02-26
Create .gitattributes

networkfusion:
## Motivation and Context

When using the dev container, it can convert all files to use linux file endings, which creates a lot of potential changes when creating a PR.
Adding a `.gitattributes` file ensures the changes are kept as windows eol.

See here for further details: https://code.visualstudio.com/docs/remote/troubleshooting#_resolving-git-line-ending-issues-in-wsl-resulting-in-many-modified-files

## Screenshots
image

## Types of changes

## Checklist


josesimoes:
Interesting and handy for sure!
Why no go with the full suggestion on the URL above?

```text
* text=auto eol=lf
*.{cmd,[cC][mM][dD]} text eol=crlf
*.{bat,[bB][aA][tT]} text eol=crlf
```

networkfusion:
> Interesting and handy for sure! Why no go with the full suggestion on the URL above?
> 
> ```
> * text=auto eol=lf
> *.{cmd,[cC][mM][dD]} text eol=crlf
> *.{bat,[bB][aA][tT]} text eol=crlf
> ```

These are un-necessary as have the expected line endings. Adding rules are only needed if we need to filter to a different line ending.

networkfusion:
With the file added:

image

Ellerbach:
> These are un-necessary as have the expected line endings. Adding rules are only needed if we need to filter to a different line ending.

Yes and not, depends how you create the files. If it doesn't hurt, go for the full suggestion! And I usually have 2 sessions of VS Code open: one with the dev container, and another one without. And yes, great finding. Looks good to me!

Ellerbach (approved):
Add the rest of the recommendation as it should not hurt and will keep things clean.
Note that if we have to run scripts on Linux, an exception should be added for those files.

networkfusion:
> Add the rest of the recommendation as it should not hurt and will keep things clean. Note that if we have to run scripts on Linux, an exception should be added for those files.

So a better tact here would be to implement:

```
* text=auto eol=crlf
*.{sh} text eol=lf
```

networkfusion:
Extra note, I think there is some confusion and I will try and explain...

By default, our repo uses Windows line endings for all files (`CRLF`). The example I posted in the PR description utilises linux line endings (`LF`).

We want to do the opposite, unless we want to do a mass repo update to switch all files.

Ellerbach:
> Unless we want to convert the whole repo to linux line endings and have CRLF only where required...

we definitelt don't want to convert all :-) So yes for the .sh soecific ending!

networkfusion:
> > Unless we want to convert the whole repo to linux line endings and have CRLF only where required...
> 
> we definitelt don't want to convert all :-) So yes for the .sh soecific ending!

Okay, done.

PR #2588, opened 2023-03-08
Adding Generic Graphic Driver implementation

Ellerbach:
## Description

- Adding Generic Graphic Driver implementation.

## Motivation and Context

- Related to https://github.com/nanoframework/Home/discussions/1231
- Following nanoframework/nanoFramework.Graphics#118.

## How Has This Been Tested?

On a real device, with real screen.

## Types of changes

## Checklist


TerryFogg:
Also, I didn't see an option for this condition, which if I hadn't needed to hack it with a current display I would not have thought it required
My Hack required an offset to the x and y for the waveshare pico-lcd1.14 display. It must be in the way it is wired up to the controller
bool DisplayDriver::SetWindow(CLR_INT16 x1, CLR_INT16 y1, CLR_INT16 x2, CLR_INT16 y2)
{

    x1 = x1 + 40;
    x2 = x2 + 40;
    y1 = y1 + 53;
    y2 = y2 + 53;

Ellerbach:
> bool DisplayDriver::SetWindow(CLR_INT16 x1, CLR_INT16 y1, CLR_INT16 x2, CLR_INT16 y2)

that is already existing in the screen class.

TerryFogg:
No, I didn't write that comment to waste your time.
 What it means is that when you choose 0 coordinate for x or y in the C# code when it gets to the controller a fixed offset is required.
The controller first visible pixel is at (40,53)

You may think this is a special case which is ok.

Ellerbach:
> The controller first visible pixel is at (40,53)

And this is where the screen properties are used. See all native drivers like this one for example: https://github.com/nanoframework/nf-interpreter/blob/deacdf3a88c3f83cc16bf8f98cf8a62195708209/src/nanoFramework.Graphics/Graphics/Displays/ILI9342_320x240_SPI.cpp#L242

And see the example of the M5STciC with the SST35S where the screen is smaller than the driver and it starts at x=52 and y=40: https://github.com/nanoframework/nanoFramework.M5Stack/blob/5e18fa4b7d95c0510ae201170da0fc4f010f7153/nanoFramework.M5StickCommon/Screen.cs#L63

I've added this feature quite some time ago. So that's the way to go to address your problem.

targets/ESP32/CMakePresets.json
```diff
@@ -308,6 +308,32 @@
                 "ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION": "50"
             }
         },
+        {
```
josesimoes, line 311: Do we really need this new preset or it just slipped into the PR?

josesimoes (approved):
LGMT! Sure looks like a nice improvement.

(I have not tested it myself)

targets/ESP32/CMakePresets.json
Ellerbach: I was even thinking adding 2 more:
- ESP32_PSRAM_REV0_GenericDisplay
- ESP32_PSRAM_REV3_GenericDisplay
Rational:
- Having those generic images available for anyone
- I guess they cover 80% of the needs
- Maybe a BLE as well

josesimoes: Well... if the plan is to have generic displays made available then we're better adding that to the existing targets that already have displays enabled. 

Otherwise I fear that so many targets start to be confusing. Not to mention that we need to maintain these on the target list, which is scattered in 3 different locations.

Ellerbach: I would love to see then at least 2 of those generic driver. It's complementary from the ones we have with dedicated drivers.
So ESP32_PSRAM_REV0_GenericDisplay and ESP32_REV0_GenericDisplay. That will cover most needs then.
I understand the maintenance but those are just the same as regular ones with graphics enabled!

josesimoes: Understood. OK, so be it. Let's have those two then.

As you are at it, please add and entry for them in the READMEs (both EN and ZN-CH).

alberk8: Looking at the Code,  Can I say that the existing native driver will still work as is but now we have a new generic SPI graphic driver?.

josesimoes (approved):
LGTM! 
The 2nd target name it's a bit extended...  😅  but... 🤷🏻‍♂️

PR #2699, opened 2023-07-18
Fix buffer overflow error on GC

kskowronsk:
## Description

- Changed minimum allocated elements from 1 to 2 as the first element is used as sentinel and HeapBlock data begin from the second one so allocating two elements ensure storage for one HeapBlock.

## Motivation and Context

- The Reachability Graph may require allocation of additional memory for MarkStack elements (storage for later HeapBlocks processing). It tries to allocate as large block as it can but when there is not enough large memory chunks it allocates smaller ones. The smallest one acceptable is for one HeapBlock element. However a single MemoryStack object uses first element as sentinel and HeapBlock elements storing begins from the second one. So the least number of allocated elements (cElement) should be 2.
- When only one (cElement ==1) is allocated the first HeapBlock is stored just after allocated memory. 

## Screenshots
Below are 3 cases of of allocations in this loop: 
`for (int cElement = g_CLR_RT_GarbageCollector.c_minimumSpaceForGC; cElement >= 1; cElement /= 2)` 
First two are correct. The third one causes memory overflow.
MarkStack allocation

## Types of changes

## Checklist

### DIFF
--- a/.devcontainer/sources/Dockerfile.All
+++ b/.devcontainer/sources/Dockerfile.All
@@ -1,4 +1,4 @@
-FROM ghcr.io/linuxcontainers/debian-slim:latest AS downloader
+FROM ubuntu:latest AS downloader
 RUN apt-get update \
     && apt-get -y install --no-install-recommends apt-utils \
     && apt-get install -y \
@@ -28,7 +28,7 @@ RUN wget $CMAKE_SCRIPT \
       && /tmp/cmake-install.sh --skip-license --prefix=/tmp/dc-extracted/cmake \
       && rm /tmp/cmake-install.sh
 
-FROM ghcr.io/linuxcontainers/debian-slim:latest AS devcontainer
+FROM ubuntu:latest AS devcontainer
 
 # Avoid warnings by switching to noninteractive
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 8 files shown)


kskowronsk:
@dotnet-policy-service agree

Ellerbach (approved):
Looks good to me

Ellerbach:
The question is now: why there is a failing test?
You can check in the pipeline, it's one of the test.
Would be great to investigate a bit the issue there.

MateuszKlatecki:
> The question is now: why there is a failing test? You can check in the pipeline, it's one of the test. Would be great to investigate a bit the issue there.

failing tests seems unrelated to this changes, see for example this Pull Request and José comment https://github.com/nanoframework/nf-interpreter/pull/2698#issuecomment-1639881872 - exacly the same situatuon

kskowronsk:
Run_UnitTests_mscorlib is failing in other pipline jobs not related to this change. Looks like a pipeline/test issue?
Look for the comment o Jose in https://github.com/nanoframework/nf-interpreter/pull/2698

josesimoes:
Yes. I can confirm that this failure has showned early this week. It's on the queue to be fixed.

josesimoes (approved):
Very interesting finding!
Lacking more testing I trust your local testing and the fix will be merged. 

As a side note: we should add stress tests to GC so we can have the GC and compaction tested  on the pipeline.

PR #2705, opened 2023-07-26
Fix computation of GC reachability for arrays with referenced objects

josesimoes:
## Description
- Add code to process array elements for referenced elements.

## Types of changes

## Checklist

### DIFF
--- a/src/CLR/Core/GarbageCollector_ComputeReachabilityGraph.cpp
+++ b/src/CLR/Core/GarbageCollector_ComputeReachabilityGraph.cpp
@@ -239,8 +239,15 @@ bool CLR_RT_GarbageCollector::ComputeReachabilityGraphForMultipleBlocks(CLR_RT_H
 
                             if (array->m_fReference)
                             {
-                                lst = (CLR_RT_HeapBlock *)array->GetFirstElement();
-                                num = array->m_numOfElements;
+                                for (uint32_t i = 0; i < array->m_numOfElements; i++)
+                                {
+                                    sub = (CLR_RT_HeapBlock *)array->GetElement(i);
+
+                                    if (sub)
+                                    {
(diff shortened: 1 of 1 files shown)


kskowronsk:
Now the algorytm does not dig into arrays and does not mark alive objects referenced by the array. As it may be the only reference to some objects it would cause these objects to be invalidly freed

josesimoes:
@kskowronsk can you please elaborate? This was tested against the GCStress test app and it run for 30 minutes without any OOM exception nor hard faults.

MateuszKlatecki:
I can confirm that changes from this PR as well from this PR https://github.com/nanoframework/nf-interpreter/pull/2706 introduce another problems. Our system is unable to even boot after making these changes.

please try run this program and check what is inside array after for loop. 

```cs
public class Program
{
    public static void Main()
    {
        const int max_array_size = 786432 - 68; //786364
        object[] arrays = new object[1047];

        for (int i = 0; i < arrays.Length; i++)
        {
            var n = max_array_size / 256;
            arrays[i] = new object[n];

            for (int j = 0; j < n; j++)
            {
                (arrays[i] as object[])[j] = new byte[2];
            }
        }
    }
}
```

I would expect our device to run out of memory (8MB of RAM) during the loop execution. After these fixes, the program comes to an end, but in the table there is:
image

kskowronsk:
The above behavior is expected for this fix as GC releases objects that are referenced by an array. 
Before this fix the array's elements were put onto the mark stack and processed later recursively.
After the fix all elements are marked in place and there is no recursion. The reachability graph ceased to "reach" into the arrays subelements so any referenced objects are not marked by GC anymore.

josesimoes:
Understood and yes, that is correct. Working on a fix.

PR #2801, opened 2023-10-25
Added Back to Top in readme.md

vivekBoii:
Screenshot from 2023-10-25 20-15-32


README.md
```diff
@@ -182,3 +183,4 @@ For more information see the [.NET Foundation Code of Conduct](https://dotnetfou
 ### .NET Foundation
 
 This project is supported by the [.NET Foundation](https://dotnetfoundation.org).
+<p align="right"><a href="#top">🔼 Back to top</a></p>
```
Ellerbach, line 186: I would prefer to have this done in pure markdown. So pointing back to the main title (which should be level 1 and not 3):

```markdown
# Welcome to the .NET **nanoFramework** interpreter repository!

// A lot of text

[🔼 Back to top](./README.md#welcome-to-the-net-nanoframework-interpreter-repository)
```

And like this, you can point in multiple places in the doc. Other advantage: it will properly work when transformed into pure HTML with DocFx for example.

PR #2803, opened 2023-10-27
Add hashing and cryptography libraries

josesimoes:
## Description
- Add both assembly declarations to CMake.
- Add implementation for STM32 with ChibiOS.
- Add implementation for ESP32.
- Add implementation for Silabs Giant Gecko in Azure RTOS.
- Update build options in several platforms/targets.
- Rework CRC32 calculator in ChibiOS nano HAL implementation.
- Regular CRC32 is now setting the config params all the time.
- Rework CRC32 calculator in Silabs Giant Gecko platform implementation.
- Move adding Mbed TLS to a binutil function.
- Add call from ChibiOS to the above.
- Add flag to Native Assembly processor to signal the requirement of Mbed TLS regardless of having it enabled through network option.

## Motivation and Context
- Addresses nanoFramework/Home#648.
- Following nanoframework/System.Security.Cryptography#1.
- Following nanoframework/System.IO.Hashing#1.

## How Has This Been Tested?
- Unit Tests in respective libraries.

## Types of changes

## Checklist

### DIFF
--- a/CMake/binutils.ChibiOS.cmake
+++ b/CMake/binutils.ChibiOS.cmake
@@ -160,6 +160,10 @@ macro(nf_add_platform_dependencies target)
 
         endif()
 
+        if(API_nanoFramework.System.Security.Cryptography)
+            FetchContent_GetProperties(mbedtls)
+        endif()
+    
         nf_add_lib_native_assemblies(
             EXTRA_INCLUDES
                 ${CHIBIOS_INCLUDE_DIRS}
@@ -171,7 +175,9 @@ macro(nf_add_platform_dependencies target)
                 ${SPIFFS_INCLUDE_DIRS}
                 ${TARGET_CHIBIOS_COMMON_INCLUDE_DIRS}
                 ${TARGET_CHIBIOS_NANOCLR_INCLUDE_DIRS}
-                ${chibios_SOURCE_DIR}/os/hal/boards/${TARGET_BOARD})
+                ${chibios_SOURCE_DIR}/os/hal/boards/${TARGET_BOARD}
+                ${mbedtls_SOURCE_DIR}/include
+                ${CMAKE_SOURCE_DIR}/src/PAL/COM/sockets/ssl/MbedTLS)
         
         add_dependencies(${target}.elf nano::NF_NativeAssemblies)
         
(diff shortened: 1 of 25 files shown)


CMake/binutils.ChibiOS.cmake
```diff
@@ -160,6 +160,10 @@ macro(nf_add_platform_dependencies target)
 
         endif()
 
+        if(API_nanoFramework.System.Security.Cryptography)
```
Ellerbach, line 163: will that work as well for small devices without any network? Typically those where you want to use a modem and still benefit of crypto elements?

targets/ESP32/_nanoCLR/nanoFramework.System.IO.Hashing/nf_sys_io_hashing_System_IO_Hashing_Crc32.cpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+    else
+    {
+        hash = ~esp_rom_crc32_le(~crc32, bufferData, bufferSize);
+    }
+
+#else
+
+    // this series does not have ROM support for CRC32
+    // for the time being this is not supported
```
Ellerbach, line 57: As some other devices may not have, I guess the next step is to build a software version for those devices, right?

Ellerbach (commented):
Great work!
My question is really about boards with not HMAC compute and how to ensure we still can do it in a software way, so no device is left behind.

CMake/binutils.ChibiOS.cmake
josesimoes: Yes, this is dependent of Mbed TLS not the network stack.
As with any other depends on a build option.

targets/ESP32/_nanoCLR/nanoFramework.System.IO.Hashing/nf_sys_io_hashing_System_IO_Hashing_Crc32.cpp
josesimoes: Yes. it's a matter of adding a lookup table for this. 
Suggest that we do than as an improvement.

Ellerbach: works for me. We can compute it as well (we have plenty of this in the IoT repo!)

Ellerbach (approved):
Thanks for this first native implementation.

PR #2805, opened 2023-11-02
Fix dataWidth check for ESP32 MMC to enable 4-bit access

Guspaz:
## Description
- Changed three locations in the ESP32 target where an enum with the possible values of 1 or 2 was being cast directly to a bool. It now compares the provided enum value to an enum entry itself.

## Motivation and Context
- The code in question is setting a flag that determines if the MMC card is mounted in 1-bit or 4-bit mode. The flag should be true for 1-bit and false for 4-bit. The functions are provided the value of an enum where 1-bit = 1 and 4-bit = 2. The functions cast either 1 or 2 to bool, which always results in true. Thus if from managed code you try to mount in 1-bit mode, then the 1-bit flag in native code is set to true. If you try to mount in 4-bit mode, the 1-bit flag is also set to true. This PR changes this from a bool cast to an equality comparison where we now check if the parameter is equal to the 1-bit enum. If it is, we set the 1-bit flag to true. If it is not (as in, the provided enum value for 4-bit), we set the 1-bit flag to false.
- The result of this change is that mounting an SD card in MMC mode results in at least 4x higher performance for file operations when in MMC mode.

## How Has This Been Tested?
- Produced the original problem on the release firmware. Mounting with MMC results in much slower performance than mounting with SPI, and performance is identical regardless of if MMC is set to 1-bit or 4-bit in managed code.
- Observed that the MMC performance issue only applied to the System.IO.FileSystem MMC support, when using the Windows.Storage MMC support, 4-bit had much faster performance than 1-bit as expected
- Built the firmware from source without any changes and replicated the problem and verified that 1-bit and 4-bit MMC mode had the same (slower than expected) performance.
- Added the fix and rebuilt the firmware. Verified that the MMC could still be mounted, but that performance was much faster (comparable to SPI mode for my particular ESP32 board) in 4-bit mode than 1-bit mode.
- I want to note that I checked the "I have tested everything locally" box in the checklist. I have tested locally, but I don't know what "all new and existing tests" would be. I validated that all my file operations work as before, and that the only difference between the before and after is the improved performance resulting from 4-bit access mode being correctly activated.

## Types of changes

## Checklist

### DIFF
--- a/targets/ESP32/_nanoCLR/System.IO.FileSystem/nf_sys_io_filesystem_nanoFramework_System_IO_FileSystem_SDCard.cpp
+++ b/targets/ESP32/_nanoCLR/System.IO.FileSystem/nf_sys_io_filesystem_nanoFramework_System_IO_FileSystem_SDCard.cpp
@@ -97,7 +97,7 @@ HRESULT Library_nf_sys_io_filesystem_nanoFramework_System_IO_FileSystem_SDCard::
     {
         case SDCard_SDInterfaceType::SDCard_SDInterfaceType_Mmc:
         {
-            bool bit1Mode = (bool)(pThis[FIELD___dataWidth].NumericByRef().s4);
+            bool bit1Mode = pThis[FIELD___dataWidth].NumericByRef().s4 == SDCard_SDDataWidth::SDCard_SDDataWidth__1_bit;
 
             int count;
             int8_t *pPins;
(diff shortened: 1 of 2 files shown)


Guspaz:
@dotnet-policy-service agree

Ellerbach (approved):
Thanks, looks good and match the discussion on Discord :-)

josesimoes (approved):
LGTM! Nice improvement. Thanks.

I took the opportunity to fix the implementation for Win.Storage

PR #2816, opened 2023-11-06
Switch dev containers to seperate directories

networkfusion:
## Description

* Removes All container.
* Adds FreeRTOS container.
* Moves containers to  seperate folders based on platform (as a start).
* Auto increments the version of containers when they are adjusted (use the tag or latest).

## Motivation and Context

They were becoming hard to manage.

## Types of changes

## Checklist


Ellerbach (commented):
Looks good. Seems quite complete now!

.devcontainer/AzureRTOS_STM32/devcontainer.json
```diff
@@ -0,0 +1,48 @@
+{
+	"name": "nanoFramework-AzureRTOS-STM32",
+	// Adjust this file to choose use of cached containers
+	// or the source docker file.
+	"image": "ghcr.io/nanoframework/dev-container-azure-rtos:latest",
```
Ellerbach, line 5: it's better to pin the version. Because once there is a change in the main container, you are noticed. When using latest you are never noticed.
Apply also on the other containers.

networkfusion: I am not quite with your thoughts there. The aim is to always be inline with the (current) toolchain. If you need to `pin` a version, that is up to the user. Is there an example you have where it should not just use the (now automated) tags?

networkfusion: Extra thoughts being that if we were to `pin` the docker image, we should also be `pinning` the commit of `nf-interpreter`

Ellerbach: just as an example, this is pinned version: 

image

If you use latest, regardless of when you'll build your container first time, it will never be updated and you'll always be sticked to it.

so it's important to pin the version and never use latest as it never update!

networkfusion:
Closing as superseeded.

PR #2820, opened 2023-11-07
Hard fault caused by unaligned Memory Access on STM32F769 series

TerryFogg:
#### SYMPTOM
If you use an STM32F7xx microcontroller with an external SDRAM,
the Cortex-M7 core may unexpectedly run into the hard fault handler because of unaligned access.
This may happen for example, when the frame buffer of an LCD, a RAM filesystem or any other data is
located into the SDRAM address range 0xC0000000 - 0xC03FFFFF (max. 4MB).
The hard fault is executed although the bit UNALIGN_TRP (bit 3) in the CCR register is not enabled.

#### CAUSE
In general, RAM accesses on Cortex-M7 based devices do not have to be aligned in any way.
The Cortex-M7 core can handle unaligned accesses by hardware.
Usually, variables should be naturally aligned because these accesses are slightly faster than unaligned
accesses.
STM32F7xx devices have the external SDRAM mapped to the address range 0xC0000000 - 0xC03FFFFF (max. 4MB).
According to the ARMv7-M Architecture Reference Manual chapter B3.1 (table B3-1), the area 0xC0000000-0xDFFFFFFF (32MB) is specified as Device Memory Type.
According to chapter A3.2.1, all accesses to Device Memory Types must be naturally aligned.
If they are not, a hard fault will execute no matter if the bit UNALIGN_TRP (bit 3) in the CCR register is
enabled or not.

## Motivation and Context
Bitmap_decoder.cpp uses memory in this area to decompress bitmaps.
This area of memory is the start of external SDRAM on the STM32F769I_Discovery.
The SDRAM starts its address is in the same address space as the 4MB designated as device memory.
Unaligned memory access faults were hit.

## How Has This Been Tested?
Bitmaps tested before were added to the Resource file which converts them to RGB565 format before 
storing them in flash.
This test was performed on byte arrays of 24bit and 8 bit which required this code to be executed.

## Types of changes

## Checklist

### DIFF
--- a/targets/ChibiOS/ST_STM32F769I_DISCOVERY/target_external_memory.c
+++ b/targets/ChibiOS/ST_STM32F769I_DISCOVERY/target_external_memory.c
@@ -69,6 +69,9 @@
 #define SDRAM_SIZE  (16 * 1024 * 1024)
 #define SDRAM_START ((void *)FSMC_Bank5_MAP_BASE)
 
+
+void SetupDeviceMemoryToEliminateUnalignedAccess();
+
 // SDRAM driver configuration structure.
 static const SDRAMConfig sdram_cfg = {
     .sdcr = (uint32_t)FMC_ColumnBits_Number_8b | FMC_RowBits_Number_12b | FMC_SDMemory_Width_32b |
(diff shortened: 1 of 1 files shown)


josesimoes (approved):
LGTM! Thanks for the fix.

PR #2849, opened 2023-12-21
Improve Storage part 2

networkfusion:
## Description

* Align names for drive defines.
* Take into account all drives when enumerating.
* Improve checks to only need a single char for drive letter.
* Rename the nanoHal header to move away from "Windows" naming.
* Remove a duplicate include.

Each change is kept seperate incase of individual merge.

## Motivation and Context

Code Maintainability.

## Types of changes

## Checklist


josesimoes (commented):
I have no problem with the rename to move away from Windows to Persistent.
Now about the defines for the drives name: that's another story.
These were using Index0...1 etc on purpose as those are the order they show when enumerated by FATFS.

Hard coding this to (for example) `USB_DRIVE_LETTER` is assuming that there will be an USB MSD on the system and that it will be at INDEX 1, which may not be the case. There can be an SD Card drive at INDEX 0. Or not... 

Any renaming seen necessary here should follow an agnostic naming pattern.

networkfusion:
> I have no problem with the rename to move away from Windows to Persistent. Now about the defines for the drives name: that's another story. These were using Index0...1 etc on purpose as those are the order they show when enumerated by FATFS.
> 
> Hard coding this to (for example) `USB_DRIVE_LETTER` is assuming that there will be an USB MSD on the system and that it will be at INDEX 1, which may not be the case. There can be an SD Card drive at INDEX 0. Or not...
> 
> Any renaming seen necessary here should follow an agnostic naming pattern.

My concept towards inprovements here are this:
We are not Windows and cannot expect a user to enumerate an environment, so the filesystem should always be constent.
FatFS does not expect an enumeration, and we are probably fighting against it.
The naming was inconsistent, We should choose one or the other.

Personally (as a developer) I (think) I would prefer a (say) 3 letter drive name that relates to the drive (like linux) rather than a 1 letter name that is based on DOS. I note that this needs us to deprecate (rather than remove).

But, would add that some of the changes in this PR are helpful and optimize code. Just need split them.

networkfusion:
@josesimoes hopefully the changes should go some way towards your comments. Let me know what else.

networkfusion:
Unfortunatly, I missed the fact that `DRIVE_LETTER_LENGTH` is used extensively with windows storage... so probably back to the drawing board.

networkfusion:
Closing as obsolete. A new PR can be created later if needed.

PR #2854, opened 2023-12-27
Improve devcontainer file structure

networkfusion:
## Description

* Moves containers to seperate folders based on platform.
Superseeds (part implementation of) #2816

## Motivation and Context

Although there are some futher changes I would like to make (in another PR), this moves the devcontainers to seperate directories which makes it easier for users to choose and develop against their respective platforms in a devcontainer.

It fixes the various issues raised against the last PR.

## How Has This Been Tested?

* Was able to select and run the devcontainers locally.
* Smoke tests still pass (using the src files).
* Publishing the images needs to be checked on when merged.
* Documentation may need to be adjusted when merged. (see here: https://github.com/nanoframework/nanoframework.github.io/tree/networkfusion-devcontainers-patch)

## Types of changes

## Checklist


networkfusion:
@alberk8 Given I know that you use devcontainers, if you can also test this and add feedback, it would be welcome.

alberk8:
@networkfusion, It works but rather annoying is that you need to select the Dev Container every time you reopen the container.  Is it possible to have something like selecting the target where it will resists across Container restart?.

.github/workflows/devcontainer-esp32.yml
```diff
@@ -23,11 +24,11 @@ jobs:
     steps:
 
     - name: Checkout Repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     - name: Get container version
       run: |
-        $dockerfileContent = Get-Content(".devcontainer/Dockerfile.ESP32")
+        $dockerfileContent = Get-Content(".devcontainer/ESP32/Dockerfile.ESP32")
```
Ellerbach, line 31: shouldn't it be with the CONTAINER_SRC_FILE?

.github/workflows/devcontainer-chibios.yaml
```diff
@@ -23,11 +24,11 @@ jobs:
     steps:
 
     - name: Checkout Repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     - name: Get container version
       run: |
-        $dockerfileContent = Get-Content(".devcontainer/Dockerfile.ChibiOS")
+        $dockerfileContent = Get-Content(".devcontainer/ChibiOS/Dockerfile.ChibiOS")
```
Ellerbach, line 31: Same here with CONTAINER_SRC_FILE and the others?

Ellerbach (commented):
hat will definitely be much cleaner! And I think the action should be adjusted, see my comment

.github/workflows/devcontainer-esp32.yml
networkfusion: @Ellerbach , I don't beleive so. It only gets the manually tagged version from the file so it can be used as the uploaded docker image tag (the same as previously). I have another PR for after this one that makes it a little more clear.

.github/workflows/devcontainer-chibios.yaml
networkfusion: See first response.

networkfusion:
@Ellerbach any response to the comments before merge?
@alberk8 I get it, but, the prompt to reopen the container at least makes sure that you are dealing with the correct container and has the benefit of making sure it is the latest version (whilst also more intuative for new users).

PR #2863, opened 2024-01-13
Add support for IPV6 sockets

AdrianSoundy:
## Description
This PR is a back port of the support for IPV6 sockets from the IDF 5.1 PR. 
Support is enabled when LWIP_LPV6 is defined in the build. This then handles both IPv4 & IPV6 ( dual stack)

Without LWIP_IPV6 defined is backwards compatible.

This PR goes with nanoframework/System.Net#309

Also added an extra ESP32 configuration for testing ( ESP32_PSRAM_REV3_IPV6). The change for enabling IPV6 is in the sdkconfig.default_rev3_ipv6.esp32 file.

May be other work required for IPv6 on other platforms.

Some more work needs to be done on the Network interface for IPV6 and Visual Studio Network IPv6 configuration. But that doesn't need to be done in this PR as this PR allows IPv6 connections.

## Motivation and Context
The IPv6 was introduced in IDF 5.1 PR as it is required for the OpenThread used by ESp32-C6 and esp32-H2
To test parts of the IDF 5.1 changes in current build to reduce risk when going to IDF 5.1
 
## How Has This Been Tested?

Using a local build of ESP32_PSRAM_REV3_IPV6.
- Connect to a network that supports IPV6
- Displayed local addresses using code
```
       IPHostEntry iph = Dns.GetHostEntry("");
       IPAddress[ ] adrs = iph.AddressList;

       Console.WriteLine($"{LH} Local interface addresses {adrs.Length}");

       foreach (IPAddress ip in adrs)
       {
           switch (ip.AddressFamily)
           {
               case AddressFamily.InterNetwork:
                   Console.WriteLine($"{LH} IPV4 Address {ip}");
                   break;
               case AddressFamily.InterNetworkV6:
                   Console.WriteLine($"{LH} IPV6 Address {ip}");
                   break;
           }
       }
```
- Create a listening socket for any address
- Connect to device using IPV4 address and IPV6 address.

Hope to put more IPv6 tests into System.Net

## Types of changes

## Checklist


josesimoes (approved):
LGTM

src/PAL/Lwip/lwIP_Sockets.cpp
```diff
@@ -315,44 +345,151 @@ SOCK_SOCKET LWIP_SOCKETS_Driver::Socket(int family, int type, int protocol)
...
+
+        case SOCK_IPPROTO_RAW:
+            protocol = IPPROTO_RAW;
+            break;
     }
 
     return lwip_socket(family, type, protocol);
 }
 
+sockaddr *Sock_SockaddrToSockaddrV4(const SOCK_sockaddr *ssa, sockaddr_in *sai, int *addrLen)
```
josesimoes, line 389: What was the reason for replacing the define with a function?
I believe it was that way to get the code "inlined" and speed up execution.
Maybe add inline attribute to hint the compiler to do that?

AdrianSoundy: It was mainly that the IPV6 complicated it so much and having it inline would add so much extra code to it. With methods we reduce code size.
I didn't think it would make any difference to performance.

PR #2871, opened 2024-01-24
CLR not starting on ESP32-C3 (XIAO_ESP32C3 firmware) when VS not connected

AdrianSoundy:
## Description
Due to task priorities the calling task(appmain) wasn't scheduled to run after starting the wire protocol and no read data was seen from the USB-Jtag connection. This stopped the CLR being started.

This is fix for ESP32-C3. It may also be relevant to other ESP32's Also we can look at reducing CPU overhead of wire protocol thread when nothing is connected. Will look at this in later PR (IDF 5.1). As this also affects esp32-C6 & H2 cpu.

## Motivation and Context
Issue reported on discord

## How Has This Been Tested?
Run esp32-C3 module with XIAO firmware and reproduce issue. All you had to do was disable device communication in VS device manager.

After change, CLR starts when VS not running.

## Types of changes

## Checklist

### DIFF
--- a/targets/ESP32/_IDF/esp32c3/app_main.c
+++ b/targets/ESP32/_IDF/esp32c3/app_main.c
@@ -44,14 +44,18 @@ void main_task(void *pvParameter)
 // Called from Esp32 IDF start up code before scheduler starts
 void app_main()
 {
+    UBaseType_t taskPriority = 5;
+
     // Switch off logging so as not to interfere with WireProtocol over Uart0
     esp_log_level_set("*", ESP_LOG_NONE);
 
     ESP_ERROR_CHECK(nvs_flash_init());
 
+    vTaskPrioritySet( NULL, taskPriority);
+
     // start receiver task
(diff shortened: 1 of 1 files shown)


josesimoes:
😯  interesting this side effect... wondering why it's happening only with C3 since this seems to be coming from the RTOS which is common... 🤔

AdrianSoundy:
@josesimoes 
Its because it uses the USB-jtag driver which does a taskyield when there is no data to read. Seems that taskYield just releases time for next ready thread to run. But not lower priority threads. When Wire protocol task is first started its seems to stop the appmain with lower priority running and starting CLR task. I'll look further into this to see if we can improve the WP for usb-jtag, see what impact it has on performance of CLR task. Maybe a small delay will work better instead of taskYield especially if VSi s not connected.

josesimoes:
Got you! Yes. That's a nasty side effect. We've diagnose this and fixed on other platforms in #2858. Didn't replace it in WP for ESP32 as we thought it would be OK leaving it with yield. Looks like that wasn't a correct assumption.

PR #2894, opened 2024-03-03
Align FatFS code page with ESP32

networkfusion:
## Description

The Code Page used by FatFS was not consistant across platforms.

## Motivation and Context

Code page `850` might be handled incorrectly.
Although the `workingBuffer` allows for it, the `stringBuffer` does not.

## Types of changes

## Checklist


josesimoes (commented):
According to wikipedia, 437 is the old legacy IBM DOS page... despite it could be the default for ESP23, we can easily override that using the corresponding config option for system.config or our own fatfs config.
Using 850 seems like a broader code page....

networkfusion:
> .

I agree. But as the ESP32 is more widely used (tested), was thinking that we should at least align with it, and then update the config if required.

networkfusion:
Additional: formatting devices with the FAT file system seems to default to `437` (I guess for backwards compatibility).

networkfusion:
Extra: http://elm-chan.org/fsw/ff/doc/appnote.html 
image

josesimoes:
Interesting all that reading! Nice finding.
Considering all the above and that we're supporting UTF-8, I would say that we should update our FF config to:

FF_CODE_PAGE 437 (recomended to decrease code size, especially when enabling Unicode)
FF_USE_LFN 3 (already there)
FF_LFN_UNICODE 2 (currently 0, this will make it support UTF-8)
FF_LFN_BUF 765 (to support UTF-8)
FF_SFN_BUF 34 (to support UTF-8)

josesimoes:
Just added the changes discussed above. Tested with ORGPAL3 and ORGPALX.

josesimoes (approved):
LGTM!

josesimoes:
Reverting the UTF-8 changes as these affect also ESP32. Should be added in another PR.

PR #2927, opened 2024-05-12
Fix ESP32 C3 out of nanoFramework Memory

alberk8:
## Description

C3 firmware cannot run beyond simple Hello world.

## Motivation and Context

The problem arise in the way the IDF 5.x calculate the available memory vs IDF 4.4.

## How Has This Been Tested?

Tested on ESP32 C3 with and without BLE.
Tested on ESP32S3 with OCT PSRAM
Tested on ESP32 REV0 and REV3

## Types of changes

## Checklist


AdrianSoundy:
This parameter is only relevant to devices without PSRAM as all remaining PSRAM goes to the managed heap.

Tests with ESP32_C3 (PSRAM not available)
When it starts it has 114K of ram available for allocation. The 100K for IDF allocations is too big and changing to 50 should give room for a TLS or BLE application.  50K for IDF gives about 62K for managed heap instead of 12K.

targets/ESP32/_nanoCLR/Memory.cpp
```diff
@@ -104,7 +104,7 @@ void HeapLocation(unsigned char *&baseAddress, unsigned int &sizeInBytes)
         size_t largestFreeBlock;
 
         // check if we have any SPIRAM
-        size_t spiramMaxSize = heap_caps_get_largest_free_block(memoryCaps | MALLOC_CAP_SPIRAM);
+        size_t spiramMaxSize = heap_caps_get_largest_free_block(MALLOC_CAP_SPIRAM);
```
AdrianSoundy, line 107: Don't see any need to change this. It just looking for max free space in SPIRAM.

AdrianSoundy (changes requested):
Changes to CMakePresets.json look good to me. Just remove change for memory.cpp as not required.

targets/ESP32/_nanoCLR/Memory.cpp
alberk8: The issue is that with `heap_caps_get_largest_free_block(memoryCaps | MALLOC_CAP_SPIRAM);` it will return the 8bit internal RAM and/or PSRAM depending if PSRAM is present.   

With `heap_caps_get_largest_free_block(MALLOC_CAP_SPIRAM); `  it is a surefire way to detect if there is PSRAM or NOT.

AdrianSoundy: Did you actually test this. As my tests show the current code returns 0 for this statement so doesn't include internal memory. The memoryCaps only indicates that the memory has to be capable of 8bit and 16bit access.
Also don't want to have to test other targets which change would possibly affect.

AdrianSoundy:
These are the memory details from C# 
 Before: (Managed heap is 12288)
 Managed:3192 Native total:271712/Free:162572/Largest:114688
 
 After: (Managed heap is 63488)
 Managed:54396 Native total:271712/Free:111364/Largest:90112
 
 You could probably reduce the ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION even further to 30 which should allow enough for secure connections

targets/ESP32/_nanoCLR/Memory.cpp
alberk8: I just tested with the following ESP32 by adding the printf below.   It did show the correct values.

```
 size_t spiramMaxSize = heap_caps_get_largest_free_block(MALLOC_CAP_SPIRAM);
 ets_printf("SPIRAM Size: %d \n", spiramMaxSize);
```
**The Result** 
ESP32 with PSRAM
SPIRAM Size: 4128768

ESP32 S3 with PSRAM
SPIRAM Size: 8257536

ESP C3
SPIRAM Size: 0

alberk8:
> These are the memory details from C# Before: (Managed heap is 12288) Managed:3192 Native total:271712/Free:162572/Largest:114688
> 
> After: (Managed heap is 63488) Managed:54396 Native total:271712/Free:111364/Largest:90112
> 
> You could probably reduce the ESP32_RESERVED_RAM_FOR_IDF_ALLOCATION even further to 30 which should allow enough for secure connections

One thing that I see in C3 is the the Native report Free is inaccurate.  Somehow it does not exclude the RAM taken by nF.

AdrianSoundy:
Memory report looks ok to me. Free memory is going down by about 50K.  The largest you can't judge as it could be a completely different heap block being reported.

targets/ESP32/_nanoCLR/Memory.cpp
AdrianSoundy: The tests you are doing above report the same values as we get now with current code. The change doesn't fix anything.
Probably both ways work.  Just didn't want to change something that isn't broken.

Also later in code we do a heap_caps_malloc with the same "memorycaps" specification.  Just keeps is consistent how it is.

alberk8: Revert change.

AdrianSoundy (approved):
Looks good

PR #2950, opened 2024-06-06
Flash Size in bootloader causing flashing issues with firmware built with IDF5.x

AdrianSoundy:
## Description

This is a issue caused by the change in IDF5.x 
Originally when flashing the NF firmware bootloader the actual flash size was updated in the bootloader header. With IDF 5.x this is not done by default as a new SHA1 digest has been added to detect corruption. 

This causes a problem when flashing a module which has a different flash size to what was specified in the build.

1. When the target flash was smaller then what was in the build we get an exception when trying to init the nvs.
2. When the target flash was larger then build size we had a problem when it tried to validate the partition sizes.  i.e. The bootloader contains size of 4MB but partition table for the 8MB flash had been loaded by nanoff because flash size is 8MB.

If built flash size was same as target flash size it all worked ok

i.e.
### ESP32_S3 firmware's built with flash size of 8MB 
- ESP32_S3 with 8MB flash = OK
- ESP32_S3 with 4MB flash = error

### ESP32  firmware's built with flash size of 4MB 
- ESP32 with 4MB flash = OK
- ESP32(Wrover) with 8MB flash = error

When target flash size is bigger then built size you could work around problem using  
_--partitiontablesize 4_  option.

Other possible ways to fix this in long term so that digest is correct.
- Updating flash size at start of application.
- Building multiple bootloaders for different flash sizes and choosing the correct bootloader in nanoff
- Patching flash size and updating digest in nanoff.

## Changes made.
Have added the option _CONFIG_ESPTOOLPY_HEADER_FLASHSIZE_UPDATE_ to all target sdkconfigs.
This will cause the flash size to be updated correctly in bootloader and all then works correctly.

**The only problem now is there is a warning in the boot log that the digest is incorrect.**

Also changed all sdkconfigs to always build with a 4MB flash size.

## Motivation and Context
- Fixes nanoFramework/Home#1492

Also Discord thread
https://discord.com/channels/478725473862549535/1247532363173330945

## How Has This Been Tested?

Tested locally with a ESP32_S3 with 4MB and 2MB psram (ESP32_S3 lolin mini )
Now flashes and runs OK

Tested a ESP32 REV3 8MB and OK

## Types of changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Introduced `CONFIG_ESPTOOLPY_HEADER_FLASHSIZE_UPDATE` configuration option across various ESP32 targets to enhance flash size management.

- **Configuration Updates**
  - Updated flash size configurations and partition table filenames for ESP32S3 targets to align with new flash size settings.


josesimoes (approved):
LGTM!
Yes, this is inline with what I've investigated when updating nanoff.
We should look into this with a long-term perspective. 
A lot of changes were brought in by IDF 5 which are making our use of it somewhat complicated and requiring a lot of adjustments and ended up on a bunch of new firmware images to cover all possible variantions and combinations...

PR #2811, opened 2023-11-06
All graphics methods now use 32-bit ARGB or RGB565 format instead of BGR

TerryFogg:
## Description
The code has been modified to only use ARGB or RGR565 format.

Code was also modified to the RotateImage to properly rotate images with a transparent background.
This was not in place, although it was already in place for stretch image.

Fixed a bug.
In testing on the STM32F769I_Discovery I ran into an exception.
Native code working on an ESP32WroverKitV4.1 would create an exception in the Bitmap_Decoder.cpp.
I eventually tracked this down to a    
ARM: STM32F7: hard fault caused by unaligned Memory Access
Included in this code is the fix for this problem as recommended by keil.

## Motivation and Context
Problems with colours. This eliminates some confusion with the formats in memory.

## How Has This Been Tested?

The modifed code was tested on
1. STM32769I_Discovery board
2. ESP32 Wrover Kit v4.1
3. ESP32 Wrover Kit 4.1 with Generic SPI class and C# class of ili9341 from the Managed Drivers.

All three tests showed the same resulting colours.
See the following testing notes below

**Testing notes**

-  A selection of 9 ARGB colours was selected to test operation

- 16-bit RGB565 values expected to be in the native bitmap for presentation to the display controller to properly send to the             display device. Examination of the memory buffer of the Bitmap showed that it was correctly in the RGB565 format stored as little endian.

- RGB565 format is commonly called "native" format in the C++/C code

- JPEG test was created using Paint.net to set the 9 pixels to the ARGB colours then saved to disk and re-read. The values re-read from the disk are "compressed" so are slightly different. The expected values were found using the eye-dropper tool of paint.net to get the hex colour values.

- Bitmaps saved in the resource file are all converted to RGB565 format before being embedded into the assembly and deployed to flash.  ( This is by design)

- The graphics code supports reading in bitmaps of 32-bit, 24-bit and 8-bit indexed,  usually from file. For this test the bitmaps are setup as a C# byte array to eliminate the need to have a file system.

- Restoring RGB565 to ARGB

  Restored results are inexact values due to the loss of 3 bits red,  2 bits green and 3 bits blue. The restoration code uses an algorithm originally from .netMF to approximate the original ARGB colour.

-   Gif values should be the same as SetPixel, GetPixel, unlike jpeg GIF uses lossless compression.

  

```csharp
// |-------------------|---------------|----------------|---------------|--------------| 
[...]

PR #2976, opened 2024-07-04
Fix broken colors and partial flush for ILI9488 display

oddbit-io:
## Description
Fixed bugs in native Ili9488 display driver:
- Corrupted colors (incorrect order BGR instead of RGB).
- Corrupted screen when using partial flush.

## How Has This Been Tested?
- Tested on LILYGO TTGO T-Relay ESP32 Wireless Module board.
- Colors tested on "Primitives" examples.
- Test application code snippet for Flush (BitBlt) provided.

## Screenshots

Left: corrupted colors, Right: fixed

ili9488fix1

Left: corrupted partial flush, Right: fixed

ili9488fix2

## Types of changes

## Checklist

BitBlt test app:
```
using System.Threading;
using System.Device.Gpio;
using nanoFramework.Hardware.Esp32;
using nanoFramework.UI;
using System.Drawing;

namespace nanoFrameworkIli9844DisplayTestApp
{
    public class Program
    {
        const int backLightPin = 32;
        const int chipSelectPin = 15;
        const int dataCommandPin = 2;
        const int resetPin = 4;
        
        static GpioController gpio = new();
        static SpiConfiguration spi = new SpiConfiguration(1, chipSelectPin, dataCommandPin, resetPin, backLightPin);
        static ScreenConfiguration screen = new ScreenConfiguration();

        public static void Main()
        {
            Configuration.SetPinFunction(19, DeviceFunction.SPI1_MISO);
            Configuration.SetPinFunction(23, DeviceFunction.SPI1_MOSI);
            Configuration.SetPinFunction(22, DeviceFunction.SPI1_CLOCK);

            DisplayControl.Initialize(spi, screen, 1024 * 1024);
            
            gpio.OpenPin(backLightPin, PinMode.Output).Write(PinValue.High);

            var buffer = DisplayControl.FullScreen;

            // test full flush
            for (int i = 0; i < 5; i++)
            {
                buffer.DrawRectangle(Color.White, 2, 0, 0, 480, 320, 0, 0,
                    Color.White, 0, 0,
                    Color.Black, 480, 320, Bitmap.OpacityOpaque);
                buffer.Flush();
                Thread.Sleep(500);

                buffer.Clear();
                buffer.Flush();
                Thread.Sleep(500);
            }

            buffer.DrawRectangle(Color.White, 2, 0, 0, 480, 320, 0, 0,
                  Color.White, 0, 0,
                  Color.Black, 480, 320, Bitmap.OpacityOpaque);

            // test partial flush
            for (int i = 0; i < 5; i++)
            {
                buffer.DrawRectangle(Color.White, 2, 100, 100, 200, 200, 0, 0,
                    Color.White, 100, 100,
                    Color.Blue, 300, 300, Bitmap.OpacityOpaque);
                buffer.Flush(100, 100, 200, 200);
                Thread.Sleep(500);

                buffer.DrawRectangle(Color.White, 2, 100, 100, 200, 200, 0, 0,
                 Color.Blue, 100, 100,
                 Color.White, 300, 300, Bitmap.OpacityOpaque);
                buffer.Flush(100, 100, 200, 200);
                Thread.Sleep(500);
            }

        }
    }
}
```

## Summary by CodeRabbit

- **Refactor**
  - Improved performance and readability of the pixel iteration logic by iterating over `x` and `y` coordinates directly, rather than using a single index.
  - Optimized the storage order of color values in the `TransferBuffer` for better efficiency.

### DIFF
--- a/src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
+++ b/src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
@@ -296,32 +296,32 @@ void DisplayDriver::BitBlt(
 
     // only 18/24 bit is supported on SPI
     for (uint32_t y = srcY; y < srcY + height; y++)
-    for (uint32_t x = srcX; x < srcX + width; x++)
-    {
-        uint32_t i = y * Attributes.Width + x;
-        
-        uint32_t element = data[i / 2]; // Each uint32 stores 2 pixels
-        uint16_t color = (i % 2 == 0) ? (element & 0xFFFF) : (element >> 16);
-
-        uint8_t b = color & 0x1F;
-        uint8_t g = (color >> 5) & 0x3F;
-        uint8_t r = (color >> 11) & 0x1F;
-
-        b = (b << 3) | (b >> 2);
-        g = (g << 2) | (g >> 4);
-        r = (r << 3) | (r >> 2);
-
-        TransferBuffer[count++] = r;
-        TransferBuffer[count++] = g;
-        TransferBuffer[count++] = b;
-
-        // can't fit another 3 bytes
-        if (count + 3 > TransferBufferSize - 1)
+        for (uint32_t x = srcX; x < srcX + width; x++)
         {
-            g_DisplayInterface.SendBytes(TransferBuffer, count);
-            count = 0;
+            uint32_t i = y * Attributes.Width + x;
+
+            uint32_t element = data[i / 2]; // Each uint32 stores 2 pixels
+            uint16_t color = (i % 2 == 0) ? (element & 0xFFFF) : (element >> 16);
+
+            uint8_t b = color & 0x1F;
+            uint8_t g = (color >> 5) & 0x3F;
+            uint8_t r = (color >> 11) & 0x1F;
+
+            b = (b << 3) | (b >> 2);
+            g = (g << 2) | (g >> 4);
+            r = (r << 3) | (r >> 2);
+
+            TransferBuffer[count++] = r;
+            TransferBuffer[count++] = g;
+            TransferBuffer[count++] = b;
+
+            // can't fit another 3 bytes
+            if (count + 3 > TransferBufferSize - 1)
+            {
+                g_DisplayInterface.SendBytes(TransferBuffer, count);
+                count = 0;
+            }
         }
-    }
     g_DisplayInterface.SendBytes(TransferBuffer, count);
     return;
 }
--- a/src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
+++ b/src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
@@ -289,15 +289,17 @@ void DisplayDriver::BitBlt(
 
     g_DisplayInterface.SendCommand(1, Memory_Write);
 
-    uint32_t numPixels = width * height;
     uint32_t count = 0;
 
     CLR_UINT8 *TransferBuffer = Attributes.TransferBuffer;
     CLR_UINT32 TransferBufferSize = Attributes.TransferBufferSize;
 
     // only 18/24 bit is supported on SPI
-    for (uint32_t i = 0; i < numPixels; i++)
+    for (uint32_t y = srcY; y < srcY + height; y++)
+    for (uint32_t x = srcX; x < srcX + width; x++)
     {
+        uint32_t i = y * Attributes.Width + x;
+        
         uint32_t element = data[i / 2]; // Each uint32 stores 2 pixels
         uint16_t color = (i % 2 == 0) ? (element & 0xFFFF) : (element >> 16);
 
@@ -309,9 +311,9 @@ void DisplayDriver::BitBlt(
         g = (g << 2) | (g >> 4);
         r = (r << 3) | (r >> 2);
 
-        TransferBuffer[count++] = b;
-        TransferBuffer[count++] = g;
         TransferBuffer[count++] = r;
+        TransferBuffer[count++] = g;
+        TransferBuffer[count++] = b;
 
         // can't fit another 3 bytes
         if (count + 3 > TransferBufferSize - 1)


andreilukichoff:
> @andreilukichoff please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@dotnet-policy-service agree

src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
```diff
@@ -289,15 +289,17 @@ void DisplayDriver::BitBlt(
...
-    uint32_t numPixels = width * height;
     uint32_t count = 0;
 
     CLR_UINT8 *TransferBuffer = Attributes.TransferBuffer;
     CLR_UINT32 TransferBufferSize = Attributes.TransferBufferSize;
 
     // only 18/24 bit is supported on SPI
-    for (uint32_t i = 0; i < numPixels; i++)
+    for (uint32_t y = srcY; y < srcY + height; y++)
+    for (uint32_t x = srcX; x < srcX + width; x++)
```
Ellerbach, line 299: please see the code-style error and PR, issue here will be fixed. We're explicit about bracket and indention.

src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
```diff
@@ -309,9 +311,9 @@ void DisplayDriver::BitBlt(
         g = (g << 2) | (g >> 4);
         r = (r << 3) | (r >> 2);
 
-        TransferBuffer[count++] = b;
-        TransferBuffer[count++] = g;
         TransferBuffer[count++] = r;
```
Ellerbach, line 314: Are you sure about the order? I remember the color order is configurable and depends on the driver.

Ellerbach (commented):
Thanks for the fix! Looks good all up, you should have a code style PR to merge.

src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
andreilukichoff: The driver only supports SPI and Ili9488 only supports 3-byte color in RGB order in SPI mode. I suppose thats why the order was hardcoded initially.

alberk8: Should we expose the RGB and BGR to the managed side so that it can be switched in the [hardware](https://github.com/DurandA/transistor-tester-lgt328p/blob/lgt328p/ILI9488.h).
```
#define RFLAG_COLOR_RGB       0b00000000     /* RGB */
#define RFLAG_COLOR_BGR       0b00001000     /* BGR */
```

andreilukichoff: If I understand correctly, this flag alters the output signal for panels with different order, while the buffer always uses RGB. If this is the case, we should substitute the hardcoded MADCTL_BGR constant in DisplayDriver::ChangeOrientation with a variable that may be either MADCTL_BGR or 0.

alberk8: Can you test it out without flipping the RGB values in code?.

andreilukichoff: I just tested it and it works exactly as I expected. It displays incorrect colors with RGB buffer and 0 value instead of MADCTL_BGR, and correct colors with MADCTL_BGR value. And vice versa with BGR buffer.
So my panel utilizes BGR pixels obviously.
I suppose the buffer should always remain RGB since there is no other option mentioned in the [datasheet](https://www.hpinfotech.ro/ILI9488.pdf). See page 122.

alberk8: Different manufacturer uses different panel with same controller.  As far as I know there are two types RGB and BGR.

andreilukichoff: Yes and controller uses RGB for buffer and flag to alter signal for different panels. I don't see why buffer should not be RGB in any case.

alberk8: May be you want to check with this [change ](https://github.com/nanoframework/nf-interpreter/pull/2811) and if that is inline with this driver.

josesimoes:
Any further developments on this?

mikmog:
This is good stuff. Made a few test runs. Performance wise it's on par with current implementation (no improvement). But colorwise it's a home run with ESP32_S3_ALL in my case.

From my (narrow) pov. Can this PR be merged? It's an isolated adjustment that only affects private(!?) builds. It fixes partial flush + colors on my display.

src/nanoFramework.Graphics/Graphics/Displays/ILI9488_480x320_SPI.cpp
```diff
@@ -289,37 +289,39 @@ void DisplayDriver::BitBlt(
...
+            b = (b << 3) | (b >> 2);
+            g = (g << 2) | (g >> 4);
+            r = (r << 3) | (r >> 2);
 
-        // can't fit another 3 bytes
-        if (count + 3 > TransferBufferSize - 1)
-        {
-            g_DisplayInterface.SendBytes(TransferBuffer, count);
-            count = 0;
+            TransferBuffer[count++] = r;
```
Ellerbach, line 314: the color scheme is also defined at initialization, what is the current initialization sequence? That maybe the broken part.

mikmog: Hmm. Might have found a way to force color order on the display during initialization. I'll give it a try

mikmog: Adjusted to RGB with a different strategy in https://github.com/nanoframework/nf-interpreter/pull/3135
Preferably the color order should come managed code. Haven't looked into that

Ellerbach: @mikmog did you find a way to have it properly set from the managed code?

mikmog: @Ellerbach Think I have, at least in theory. 

One way to do it is to implement the display driver Orientation's native side for the ILI9-drivers. Currently only the generic driver have them implemented in native, it seems. 

That would allow us to simply set MADCTL_BGR/MADCTL_RGB on the managed side like below.
https://github.com/nanoframework/nanoFramework.Graphics/blob/c8a8d911e388887a6e65bf2218b59e1d6c196e98/ManagedDrivers/Ili9341/Ili9341.cs#L145

However I would rather have the generic display driver support the ILI9-displays, and extend the managed drivers with a definition for ILI9488. Have ambitions to take a look at it. Just don't know when 😥 🤷

Ellerbach: yes, when I implemented the generic driver, I reused elements from each drivers, so it basically have more features than the others. And moved the implemented features into the managed drivers, so there was the same features from native to managed. It's then indeed fairly easy to add features on the managed rather than the native side indeed.
Hapy you're still looking at all this :-)

PR #3023, opened 2024-09-23
Metadata of native assemblies is now generated and exposed in virtual device

frobijn:
## Description

- A file native_assemblies.csv is generated when a firmware package is built. This is done by CMake that reads the data in the source files. The native_assemblies.csv file lists name, version and checksum for the included native assemblies.

- New option --getnativeassemblies added to nanoclr.exe (and corresponding code to the native CLR host). This produces a list of name, version and checksum for the native assemblies included in the runtime embedded in nanoclr.exe or external file. Can still be used with "old" runtimes.
- Rename `--localinstance` to `--clrpath` and made it common to reuse on all verbs.

It would be great if the latest version of each firmware package on cloudsmith would be re-created (with the same version number) so that the `native_assemblies.csv` files are part of the packages. I have no idea how to make that happen.

## Motivation and Context

This is a PR related to the "controlled update" strategy as alternative for the current nanoFramework's "auto-update" strategy. It is required for the "controlled update" strategy to work, but there are no other PR dependent on this PR. All other software changes can be made and tested independent of this PR.

For a description of how the "controlled update" strategy would work, see the [concept documentation](https://github.com/frobijn/nF_nanoframework.github.io/blob/Versioning/content/getting-started-guides/getting-started-versioning.md) in my fork.

For a description of all related changes required to make the  "controlled update" strategy work, see a [detailed description](https://github.com/frobijn/nF_nanoFirmwareFlasher/blob/Versioning_Development/Changes_Versioning.md) in another fork.

## Long description

The changes in this repository have been made to get a list of names and versions of the native assemblies that are used in a runtime, without having to connect to a device via a serial port and retrieve the list via the wire protocol.

- NanoCLR.exe: an option has been added: --getnativeassemblies, similar to (and can be combined with) --getversion. The result is a list of names, version and checksums sent to the standard output.

    It calls a method in the CLR host assembly to get information about the native assemblies (a .NET class). The data for that class is deserialized from data obtained via two additional methods in the native NanoCLR runtime that serializes the static constants in the runtime with native assembly information. If a tool requires the list of native assemblies, it can run the nanoclr.exe and parse its output.

- NanoCLR.exe: added --clrinstancepath to use with --getnativeassemblies, so that the above keeps working if nanoclr.exe is updated but an older runtime is used. If the instance does not support the two additional methods in the native NanoCLR runtime, no details about the native assemblies is displayed, but instead a message (if verbosity is high enough).

- NanoCLR.exe: changes have been tested by running the tool, for the "embedded" runtime, the same runtime via --clrinstancepath, and the current version of the runtime (from Cloudsmith) via --clrinstancepath.

- CMake/Modules/FindNF_NativeAssemblies.cmake: create a `native_assemblies.csv` file in the build output directory of a firmware package that contains the names, version and checksums of the native assemblies selected to be part of the firmware.

    - The CMake code to extract version and checksum works well for the native assemblies that use the skeleton generated by the metadata processor. There are two native assemblies that define the checksum in a different way:
        - src\nanoFramework.Runtime.Events\nf_rt_events_native_nanoFramework_Runtime_Events_EventSink.cpp has a g_CLR_AssemblyNative_nanoFramework_Runtime_Events_EventSink_DriverProcs with a #define as checksum (DRIVER_INTERRUPT_METHODS_CHECKSUM). However, this is not the checksum of the native assembly corresponding to the nanoFramework.Runtime.Events .NET class library - that is located in src\nanoFramework.Runtime.Events\nf_rt_events_native.cpp. This exception can be ignored for our purpose.
        - src\CLR\CorLib\corlib_native.cpp: The checksum depends on NANOCLR_REFLECTION that corresponds to TARGET_NANOCLR_REFLECTION in CMake.As the corlib is already treated as a special (as the corlib is always included and treated differently in CMake), the parsing is also done differently.

    - File creation has been tested for a few firmware packages. As there does not seem to be a target with `NF_FEATURE_SUPPORT_REFLECTION = OFF`, that part of the CMake script could not be tested. The CMake regex has been tested by temporarily using it for the `NF_FEATURE_SUPPORT_REFLECTION = ON` case.
    - As .csv files are copied to the firmware .zip package, it is assumed the `native_assemblies.csv` will become part of the firmware package. This is not tested.

- Added .editorconfig.

## How Has This Been Tested?

See long description.

## Types of changes

## Checklist

## Release Notes

- **New Features**
	- Introduced a new `.editorconfig` file to enforce consistent coding styles across the project.
	- Added properties to `ClrInstanceOperationsOptions` for retrieving native assembly details and specifying local instances.
	- New methods for obtaining native assembly counts and information have been added to enhance interop functionality.
	- A new class `NativeAssemblyDetails` has been introduced to encapsulate information about native assemblies.
	- Added functionality to create a CSV file containing native assembly versions.

- **Bug Fixes**
	- Improved error handling for local CLR instance validation.

- **Documentation**
	- Updated licensing comments to reflect MIT licensing across various files.
	- Added a spelling exclusion dictionary to improve spell-check accuracy.

- **Chores**
	- Introduced a spelling exclusion dictionary to improve spell-check accuracy.


targets/netcore/nanoFramework.nanoCLR/nanoCLR_native.cpp
```diff
@@ -182,3 +182,36 @@ const char *nanoCLR_GetVersion()
...
+        memcpy((void *)data, &g_CLR_InteropAssembliesNativeData[i]->m_Version.iBuildNumber, sizeof(CLR_UINT16));
+        data += sizeof(CLR_UINT16);
+
+        memcpy((void *)data, &g_CLR_InteropAssembliesNativeData[i]->m_Version.iRevisionNumber, sizeof(CLR_UINT16));
+        data += sizeof(CLR_UINT16);
+
+        hal_strcpy_s((char *)data, 128, g_CLR_InteropAssembliesNativeData[i]->m_szAssemblyName);
+        data += 128;
+    }
+}
```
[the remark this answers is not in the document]
frobijn: Suggestion applied.

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            return result;
+        }
+
+        /// <summary>
+        /// Deserialize the description of the native assembly.
+        /// In nf-debugger this is done via a converter. The code in this
+        /// constructor is essentially the same.
+        /// </summary>
+        /// <param name="reader"></param>
```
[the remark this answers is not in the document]
frobijn: If the AI was really intelligent, the rabbit would know that in this case code duplication of an auxiliary method is better than making nanoclr dependent on a library that has no added value for this project

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,119 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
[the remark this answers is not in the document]
frobijn: This is outdated, BOM removed.

.editorconfig
```diff
@@ -0,0 +1,213 @@
...
+# avoid this. unless absolutely necessary
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
+
+# Types: use keywords instead of BCL types, and permit var only when the type is clear
+csharp_style_var_for_built_in_types = false:suggestion
+csharp_style_var_when_type_is_apparent = false:none
+csharp_style_var_elsewhere = false:suggestion
```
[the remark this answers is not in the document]
frobijn: This is the .editorconfig from the .NET core team, I won't touch it.

.editorconfig
```diff
@@ -0,0 +1,213 @@
...
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+# Generated code
+[*{_AssemblyInfo.cs,.notsupported.cs,AsmOffsets.cs}]
```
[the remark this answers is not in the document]
frobijn: This is the .editorconfig from the .NET core team, I won't touch it.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -62,6 +62,53 @@ option(API_Hardware.GiantGecko                          "option for Hardware.Gia
...
+    foreach(apiSourceFile ${nativeAssemblySources})
+
+        file(READ "${apiSourceFile}" sourceCode)
+
+        string(REGEX MATCH "[ \t\r\n]+CLR_RT_NativeAssemblyData[ \t\r\n]+g_CLR_AssemblyNative_${apiNamespaceWithoutDots}[ \t\r\n]*=[^{]+{[^0]+0x([0-9A-Za-z]+)[^{]+{[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)"  _ "${sourceCode}")
+
+        if(NOT "${CMAKE_MATCH_1}" STREQUAL "")
+            list(APPEND NF_NativeAssemblies_VERSIONS "${apiNamespace},${CMAKE_MATCH_2}.${CMAKE_MATCH_3}.${CMAKE_MATCH_4}.${CMAKE_MATCH_5},${CMAKE_MATCH_1}")
+            break()
+        endif()
```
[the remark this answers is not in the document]
frobijn: That may be a valid addition, but:

- The error would be: the format of the version information is not in the expected format. That seems to be highly unlikely as it seems to be generated.
- To be sure it is *always* an error, the thing to investigate is: will it fail for native components that are not related to a managed component (=> error is generated but there is no need for the native assembly metadata to be added to the list). Not sure about that.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -62,6 +62,53 @@ option(API_Hardware.GiantGecko                          "option for Hardware.Gia
...
+        if(NOT "${CMAKE_MATCH_1}" STREQUAL "")
+            if (NF_FEATURE_SUPPORT_REFLECTION)
+                list(APPEND NF_NativeAssemblies_VERSIONS "${apiNamespace},${CMAKE_MATCH_3}.${CMAKE_MATCH_4}.${CMAKE_MATCH_5}.${CMAKE_MATCH_6},${CMAKE_MATCH_1}")
+            else()
+                list(APPEND NF_NativeAssemblies_VERSIONS "${apiNamespace},${CMAKE_MATCH_3}.${CMAKE_MATCH_4}.${CMAKE_MATCH_5}.${CMAKE_MATCH_6},${CMAKE_MATCH_2}")
+            endif()
+
+            break()
+        endif()
+
```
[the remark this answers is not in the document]
frobijn: Ah, the AI is not smart enough to see that this will fail only if someone messes with the code in the *.cpp file, and there is a huge warning in that file to *not* change the code or else modify this CMake file.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -62,6 +62,53 @@ option(API_Hardware.GiantGecko                          "option for Hardware.Gia
...
+    endforeach()
+
+endmacro()
```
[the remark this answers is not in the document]
frobijn: An AI should know that two macros with similar code but different logic should be preferred to one quite complex macro  that is much harder to comprehend.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -123,6 +172,8 @@
     list(APPEND NF_NativeAssemblies_SOURCES "${${interopAssemblyName}_SOURCES}")
     list(REMOVE_DUPLICATES NF_NativeAssemblies_SOURCES)
 
+    # add the assembly version to the list
+    AddNativeAssemblyVersion(interopAssemblyName interopAssemblyNameWithoutDots "${${interopAssemblyName}_SOURCES}")
```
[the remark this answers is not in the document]
frobijn: Indeed. Corrected

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -424,6 +483,10 @@
 # ... now add Interop Assemblies table to ChibiOS nanoCLR sources list
 list(APPEND NF_NativeAssemblies_SOURCES "${CMAKE_CURRENT_BINARY_DIR}/CLR_RT_InteropAssembliesTable.cpp")
 
+# create a .csv file with native assembly versions in the output directory
+string(REPLACE ";" "\r\n" NF_NativeAssemblies_CSV "${NF_NativeAssemblies_VERSIONS}")
+file(WRITE "${CMAKE_BINARY_DIR}/native_assemblies.csv" "${NF_NativeAssemblies_CSV}")
+
```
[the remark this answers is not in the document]
frobijn: Suggestion applied

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -62,6 +62,53 @@
...
+        string(REGEX MATCH "[ \t\r\n]+CLR_RT_NativeAssemblyData[ \t\r\n]+g_CLR_AssemblyNative_${apiNamespaceWithoutDots}[ \t\r\n]*=[^{]+{[^0]+0x([0-9A-Za-z]+)[^{]+{[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)"  _ "${sourceCode}")
+
+        if(NOT "${CMAKE_MATCH_1}" STREQUAL "")
+            list(APPEND NF_NativeAssemblies_VERSIONS "${apiNamespace},${CMAKE_MATCH_2}.${CMAKE_MATCH_3}.${CMAKE_MATCH_4}.${CMAKE_MATCH_5},${CMAKE_MATCH_1}")
+            break()
+        endif()
+
+    endforeach()
+
+endmacro()
```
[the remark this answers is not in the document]
frobijn: No.

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,119 @@
...
+                return null;
+            }
+            if (numAssemblies == 0)
+            {
+                return result;
+            }
+
+            byte[] data = new byte[numAssemblies * /* size per assembly: */  (4 + 4 * 2 + 128 * 1)];
+            Interop.nanoCLR.nanoCLR_GetNativeAssemblyInformation(data, data.Length);
+
```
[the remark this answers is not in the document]
frobijn: That really is a very theoretical situation.

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,119 @@
...
+            using var buffer = new MemoryStream(data);
+            using var reader = new BinaryReader(buffer);
+            for (int i = 0; i < numAssemblies; i++)
+            {
+                result.Add(new NativeAssemblyDetails(reader));
+            }
+
```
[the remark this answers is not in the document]
frobijn: That is really really unlikely.

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,122 @@
...
+namespace nanoFramework.nanoCLR.Host
+{
+    /// <summary>
+    /// Information about a native assembly.
+    /// </summary>
+    public sealed class NativeAssemblyDetails
+    {
+        #region Properties
+        /// <summary>
+        /// Name of the assembly.
```
Ellerbach, line 18: Minor: can you please use the standard way for the comments for the properties? So, for properties, it should start by "Gets or sets" or in this case "Gets". I know it's still not there yet everywhere but at least for new code, it's better to start clean :-)

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,122 @@
...
+        private static string GetZeroTerminatedString(byte[] buf, bool fUTF8)
+        {
+            if (buf == null) return null;
+
+            int len = 0;
+            int num = buf.Length;
+
+            while (len < num && buf[len] != 0) len++;
+
+            if (fUTF8) return Encoding.UTF8.GetString(buf, 0, len);
```
Ellerbach, line 117: we are using as a code style explicit brackets for all the if, for, etc.. So proper way for the code will be:

```csharp
if (fUTF8) 
{
    return Encoding.UTF8.GetString(buf, 0, len);
}
else 
{
    return Encoding.ASCII.GetString(buf, 0, len);
}
```

Other more "modern" way:

```csharp
return fUTF8 ? Encoding.UTF8.GetString(buf, 0, len) : Encoding.ASCII.GetString(buf, 0, len);
```

If you prefer condenced, then second choice is better ;-)

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="fUTF8"></param>
+        /// <returns></returns>
+        private static string GetZeroTerminatedString(byte[] buf, bool fUTF8)
+        {
+            if (buf == null) return null;
+
+            int len = 0;
+            int num = buf.Length;
+
+            while (len < num && buf[len] != 0) len++;
```
Ellerbach, line 115:
```suggestion
            while (len < num && buf[len] != 0) 
            {
                len++;
            }
```

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
```diff
@@ -0,0 +1,122 @@
...
+            ushort numAssemblies;
+            try
+            {
+                numAssemblies = Interop.nanoCLR.nanoCLR_GetNativeAssemblyCount();
+            }
+            catch (EntryPointNotFoundException)
+            {
+                return null;
+            }
+            if (numAssemblies == 0)
```
Ellerbach, line 61:
```suggestion

            if (numAssemblies == 0)
```

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
```diff
@@ -93,6 +91,9 @@ public nanoCLRHostBuilder UseWireProtocolPort(IPort port)
         public string GetCLRVersion() =>
             Interop.nanoCLR.nanoCLR_GetVersion();
 
+        public List<NativeAssemblyDetails>? GetNativeAssemblies()
```
Ellerbach, line 94:
```suggestion
        public List<NativeAssemblyDetails>? GetNativeAssemblies() =>
```

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
```diff
@@ -93,6 +91,9 @@ public nanoCLRHostBuilder UseWireProtocolPort(IPort port)
...
+            => NativeAssemblyDetails.Get();
```
Ellerbach, line 95:
```suggestion
            NativeAssemblyDetails.Get();
```

targets/netcore/nanoFramework.nanoCLR/nanoCLR_native.cpp
```diff
@@ -182,3 +182,44 @@ const char *nanoCLR_GetVersion()
...
+        return false; // Buffer too small
+    }
+
+    // clear buffer memory
+    memset((void *)data, 0, size);
+
+    // fill the array
+    for (uint32_t i = 0; i < g_CLR_InteropAssembliesCount; i++)
+    {
+        memcpy((void *)data, &g_CLR_InteropAssembliesNativeData[i]->m_checkSum, sizeof(uint32_t));
```
Ellerbach, line 205: rather than uint32_t CLR_UINT32 for consistency?  (not 100%, just asking)

Ellerbach (commented):
Thanks, couple of comments, mainly on the code style. I'll let José comment more about the rest. The implementation still looks good to me and aligned with the rest.

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
frobijn: Yes, this is a verbatim copy from the code in nf-debugger, don't blame me :-)

targets/netcore/nanoFramework.nanoCLR/nanoCLR_native.cpp
frobijn: No particular reason, will use CLR_UINT32.

targets/netcore/nanoFramework.nanoCLR.Host/NativeAssemblyDetails.cs
frobijn: You'll find this a lot in my code. If the next line of code after a } clearly "belongs" to the part within {}, there is no blank line. That is: if there would not be a blank line between the {..} and the next statement if the {} were not needed for syntactical reasons, then the blank line is not present if {} is needed.

It is weird if you do not use a blank line in:
```
uint count = arg is null ? 0 : arg.Length;
if (count == 0) ...
```
But you do in:
```
uint count;
if (arg is null) {
    count = 0;
}
else
{
    count = arg.Length;
}

if (count == 0) ...
```

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
frobijn: Are you sure you want to enforce this?

I always put the arrow on the next line, unless the expression is really short. I've picked up the style (end of line before arrow) from internet examples; the internet seems to be swamped with that.

There doesn't seem to be an .editorconfig style for this aspect of the formatting (Resharper has one), so it is not easy to make sure that this code style rule is satisfied.

josesimoes:
Fixed the description sections to follow PR template.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -62,6 +62,53 @@ option(API_Hardware.GiantGecko                          "option for Hardware.Gia
...
+    # find the source file that contains the value of CLR_RT_NativeAssemblyData g_CLR_AssemblyNative_${apiNamespaceWithoutDots}
+    foreach(apiSourceFile ${nativeAssemblySources})
+
+        file(READ "${apiSourceFile}" sourceCode)
+
+        string(REGEX MATCH "[ \t\r\n]+CLR_RT_NativeAssemblyData[ \t\r\n]+g_CLR_AssemblyNative_${apiNamespaceWithoutDots}[ \t\r\n]*=[^{]+{[^0]+0x([0-9A-Za-z]+)[^0]+0x([0-9A-Za-z]+)[^{]+{[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)[ \t\r\n]*,[ \t\r\n]*([0-9]+)"  _ "${sourceCode}")
+
+        if(NOT "${CMAKE_MATCH_1}" STREQUAL "")
+            if (NF_FEATURE_SUPPORT_REFLECTION)
+                list(APPEND NF_NativeAssemblies_VERSIONS "${apiNamespace},${CMAKE_MATCH_3}.${CMAKE_MATCH_4}.${CMAKE_MATCH_5}.${CMAKE_MATCH_6},${CMAKE_MATCH_1}")
```
josesimoes, line 99: Please add the trailing `0x` to the checksum to be perfectly clear that a value in hexa.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -62,6 +62,53 @@ option(API_Hardware.GiantGecko                          "option for Hardware.Gia
...
+            else()
+                list(APPEND NF_NativeAssemblies_VERSIONS "${apiNamespace},${CMAKE_MATCH_3}.${CMAKE_MATCH_4}.${CMAKE_MATCH_5}.${CMAKE_MATCH_6},${CMAKE_MATCH_2}")
```
josesimoes, line 101: Same here.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -424,6 +483,13 @@ configure_file("${CMAKE_SOURCE_DIR}/InteropAssemblies/CLR_RT_InteropAssembliesTa
 # ... now add Interop Assemblies table to ChibiOS nanoCLR sources list
 list(APPEND NF_NativeAssemblies_SOURCES "${CMAKE_CURRENT_BINARY_DIR}/CLR_RT_InteropAssembliesTable.cpp")
 
+# create a .csv file with native assembly versions in the output directory
+string(REPLACE ";" "\r\n" NF_NativeAssemblies_CSV "${NF_NativeAssemblies_VERSIONS}")
+file(WRITE "${CMAKE_BINARY_DIR}/native_assemblies.csv" "${NF_NativeAssemblies_CSV}" RESULT_VARIABLE WRITE_RESULT)
```
josesimoes, line 488: For some reason the last line in all generated files is always showing as:
`System.Device.UsbStream,100.0.0.6,21BF24CBRESULT_VARIABLEWRITE_RESULT` ... 🤔 

All the others are correct.

frobijn:
> Fixed the description sections to follow PR template.

What was the problem? I stared at it for some time but didn't see what the issue was.

targets/netcore/nanoFramework.nanoCLR.CLI/ClrInstanceOperationsOptions.cs
```diff
@@ -38,6 +42,14 @@ public class ClrInstanceOperationsOptions
             HelpText = "Specify a version of nanoCRL to install.")]
         public string TargetVersion { get; set; }
 
+        [Option(
```
josesimoes, line 45: This is a duplication of `ExecuteCommandLineOptions`.LocalInstance`, isn't it?

josesimoes (commented):
I'm not a big fan of the new option to list the assemblies... 🤔 
This is a deviation from all the other CLRs (on which this data is already possible to request) which makes it necessary to process it differently _when_ connected to the virtual device and _when_ connected to hardware.

Seems like a very very niche usage scenario that complicates the usage of, just because it deviates...

targets/netcore/nanoFramework.nanoCLR.CLI/ClrInstanceOperationsOptions.cs
frobijn: Yes and no. It serves the same purpose. But the way the options are organised is per "verb", so per "instance" and "run", so a new option has to be introduced.

My choice was to name it differently than the LocalInstance option for "run", as the LocalInstance is quite tricky. The LocalInstance  option is described as the specification of a *file* that should be used as runtime. But if you try that as a simple user and pass "<path>\nanoclr-1.2.3.dll", that won't work. Nanoclr uses the *directory* of the path and always uses the file name "nanoFramework.nanoCLR.dll". There are good reasons for that if you look at the code, but as a user (at least for me) it was a surprise. Because of backward compatibility the existing LocalInstance should not be changed, but I decided to give the new option a better name, let the user pass the *directory* and mention in the description that the file name is fixed.

CMake/Modules/FindNF_NativeAssemblies.cmake
frobijn: The RESULT_VARIABLE WRITE_RESULT was a coderabbitai recommendation, as protection against write errors. It looked good and RESULT_VARIABLE does exist for file(), but apparently not in combination with file (WRITE). I didn't catch that, this is the first time I work with CMake... Fixed it.

So apparently coderabbitai is a true AI, it makes up its own syntax.

CMake/Modules/FindNF_NativeAssemblies.cmake
frobijn: Done in next commit. Also added 0x to the output of nanoclr.exe.

targets/netcore/nanoFramework.nanoCLR.CLI/ClrInstanceOperationsOptions.cs
josesimoes: Understood on the potential confusion about the path + name. That can be easily fixed, of course.
Now about the verbs, those are organized in that manner on purpose.

The CLI at the end of the day acts as a loader for a CLR instance. That's why you find the `localinstance` option under the _run_ verb. Because one is requesting the tool to RUN _that_ CLR instance (_in lieu_ of the default one).

You're proposing to move this to the INSTANCE verb. That is not logical because, at that point of the execution its already _running_  THAT instance that was previously requested. All this works perfectly garbled and mixed anyway considering that the CLI options are parsed and only after that the thing will actually do something.
Still, I would prefer to keep the consistency and have the options where they belong.

All for improving the option help text, documentation and code that processes it to simplify.

No strong opinion on allowing a DLL name different than `nanoFramework.nanoCLR.dll`... if that seems important... this was more a like an extra validation and using the default name seemed like a good approach to make the user life easier.

frobijn:
> I'm not a big fan of the new option to list the assemblies... 🤔 

Actually it is worse than that. There are three ways a runtime is packaged:

- for all real hardware: as a zip file, published to cloudsmith. This is the nicest way, as it was easy to add a .csv file to the package. Once you have the package, you have almost all information. Only the name of the platform is nowhere to be found in the package.

- for the virtual device: as part of nanoclr.exe. The only way to get information on the runtime is by executing nanoclr.exe and reading the standard output. The alternative (producing a .csv file) is actually worse, because then the code that is running nanoclr.exe must find a directory to store the file, rather than reading and parsing stdout.

- for the virtual device: as a dll (WIN_DLL_nanoCLR), published to cloudsmith. If the dll is downloaded, you cannot use a name like WIN_DLL_nanoCLR-1.12.0.74.dll so you must store it as, e.g., "WIN_DLL_nanoCLR-1.12.0.74/nanoFramework.nanoCLR.dll", otherwise you cannot use the DLL in nanoclr.exe. And if you want to know the native assemblies, you have to run "nanoclr.exe instance --clrinstancepath WIN_DLL_nanoCLR-1.12.0.74 --getnativeassemblies".

The last version is hidden in an auto-update scenario, as it is only used by "nanoclr instance --update". But if you want to freeze the framework version, I think you need to make a backup of that as well.

On the upside: all this information is required for tooling that runs at build-time. Once that tooling is part of the framework, I don't think that there are (m)any users that have a use case for retrieving the native assemblies. Then a new command line option that probably no human uses doesn't seem to be a bad solution.

targets/netcore/nanoFramework.nanoCLR.CLI/ClrInstanceOperationsOptions.cs
frobijn: > You're proposing to move this to the INSTANCE verb

No, not to move. Both verbs need this option:

 - If you have a runtime (dll) and want to know the native assemblies, there is no other way than to use "nanoclr.exe instance" that already provides information about the runtime and has the option to update the runtime. Either that or you create a separate tool for that.
 
 - If you want to use the runtime to run assemblies, you'd use the "nanoclr.exe run". This is also used by the test framework.
 
 The way the CLI program is organized, there is no way to have common options. So I think the choices are: use the same name for both verbs (to stress that it is both concern the runtime DLL) or allow different names and use that to make one option a bit more clear.
 
 > No strong opinion on allowing a DLL name different than nanoFramework.nanoCLR.dll..

Maybe I don't understand how the various components work together, but I thought that the runtime DLL is the same one that is referred to by the various `DLLImport` declarations in the managed nanoFramework.nanoCLR.Host assembly. AFAIK if the DLL is a native DLL, the name used in the DLLImport must match the name of the DLL.

BTW I don't *want* to use a different file name, but the option suggests that it is possible. Put me on the wrong track.

josesimoes: Absolutely correct on that last part (I overlooked that) it must have that name because of the DLL import declarations as you've pointed out.

josesimoes:
@frobijn I've made some improvements on the code and pushed to the PR.

- Refactor verb options to remove duplicate ones and improved the description to address your valid concern about clarity,
- Add new exit code to warn about mutually exclusive options (can't use the same as missing DLL as that can lead to confusion).
- Tidy up output of native assemblies list. Since we code for humans, not machines, better have it nicely presented 😉 
Here's the output:
image

Please review!

targets/netcore/nanoFramework.nanoCLR.CLI/ExecuteCommandLineOptions.cs
```diff
@@ -11,7 +11,7 @@ namespace nanoFramework.nanoCLR.CLI
     [Verb(
         "run",
         HelpText = "Run nanoCLR assembly.")]
-    public class ExecuteCommandLineOptions
+    public class ExecuteCommandLineOptions : CommonOptions
```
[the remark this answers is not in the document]
josesimoes: it does...

targets/netcore/nanoFramework.nanoCLR/nanoCLR_native.cpp
```diff
@@ -182,3 +182,50 @@ const char *nanoCLR_GetVersion()
...
+
+        memcpy(data, &g_CLR_InteropAssembliesNativeData[i]->m_Version.iRevisionNumber, sizeof(CLR_UINT16));
+        data += sizeof(CLR_UINT16);
+
+        hal_strcpy_s((char *)data, 128, g_CLR_InteropAssembliesNativeData[i]->m_szAssemblyName);
+        data += 128;
+    }
+
+    return true; // Success
+}
```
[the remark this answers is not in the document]
josesimoes: Understood on the suggestion about ASSEMBLY_NAME_SIZE.
Will keep that in mind for a future code change.

targets/netcore/nanoFramework.nanoCLR.Host/NanoClrHostBuilder.cs
Ellerbach: well, it's about consistency. I'm not focused on one or the other but rather having consistency.

frobijn: Hmmm. Consistency has many aspects that really make a difference. Legal (file header). Coding, e.g., "from x in list where x == 5..." vs "list.Where (x => x == 5)". Coding style for maintainability, e.g., "var" vs explicit type. For readability, e.g., curly braces on new lines vs obfuscated code style. 

And then there are coding style aspects that concern vanity: everyone has a different preference, but if the same code is formatted in a different way, most people will not even notice and will certainly  understand the code equally well. Like whether a space should be inserted before or after an opening parenthesis. Thankfully we now have auto-format and .editorconfig, because life is too short to spend any time on these coding style aspects, as this type of consistency has no other benefit.

Unfortunately the auto-format/.editorconfig does not support any rules yet. I'll commit all your suggestions, but be warned that there may be many more in future PR's :-)

frobijn:
Nice presentation! I've added a sort by name to make it even more readable.

Checked the replacement of the LocalInstance by PathToClrInstance option (few changes) and tested it. Works well.

Two nf-interpreter checks seem to fail for the commit. I have no clue what's wrong or how to fix that.

josesimoes (approved):
LGTM! Nice improvement. 👍🏻

josesimoes:
@frobijn to answer your question above: there is no easy way to recreate the published fw pacakges. That would require a lot of manual work. From now on that will be there.

PR #3174, opened 2025-06-01
Logic fix

frobijn:
## Description
Corrected incorrectly inverted logic.

## Motivation and Context

I've stumbles on this while investigating another issue.

The logic on lines 125+ are inverted from lines 60+, but the check for hidden/system attributes was not inverted correctly:

- when looking for directories, the old condition on line 125 was false (and would include in the result although there was no room allocated for it) if the directory entry is a hidden or system file or directory.

- when looking for files, the old condition on line 154 would skip only files that are hidden and system, instead of hidden or system (that is the condition on line 83).

## How Has This Been Tested?

I've checked the code in a custom firmware - it compiles and has the same results for other than the described cases. I haven't tried to create a test with the above cases.

## Types of changes

## Checklist


josesimoes (approved):
LGTM! Nice finding.

PR #3361, opened 2026-05-19
ESP32_C61/C5 device map updates and I2C change

AdrianSoundy:
## Description

- Changed the default pins used for I2C1 bus as these conflicted with psram/flash and caused crashes.
- Add extra validation of pin numbers to I2c so a normal exception is given if pin isn't assigned or an invalid/reserved pin
- Removed the extra cmake definitions for C61 
- Updated tasks.template.json as command lines seem to be for older esptool
 

## Motivation and Context
- Further testing on I2C for C61
- Fix crashes from using wrong pins during testing

## How Has This Been Tested?
Tested locally with SDD1306 display samples

## Types of changes

## Checklist


.vscode/tasks.TEMPLATE.json
```diff
@@ -18,7 +18,7 @@
     {
       "label": "Flash nanoCLR to ESP32 | ESP32-S2",
       "type": "shell",
-      "command": "python ${env:IDF_PATH}/components/esptool_py/esptool/esptool.py --chip auto --port \"${input:comPort}\" --baud 1500000 --before \"default_reset\" --after \"hard-reset\" write-flash -z --flash-size detect 0x1000 ${workspaceFolder}/build/bootloader/bootloader.bin 0x10000 ${workspaceFolder}/build/nanoCLR.bin 0x8000 ${workspaceFolder}/build/partitions_${input:esp32Partitions}.bin",
+      "command": "python ${env:IDF_PATH}/components/esptool_py/esptool/esptool.py --chip auto --port \"${input:comPort}\" --baud 1500000 --before \"default_reset\" --after \"hard_reset\" write_flash -z --force --flash_size detect 0x1000 ${workspaceFolder}/build/bootloader/bootloader.bin 0x10000 ${workspaceFolder}/build/nanoCLR.bin 0x8000 ${workspaceFolder}/build/partitions_${input:esp32Partitions}.bin",
```
josesimoes: The underscore options were from old esptool... in the latest versions those were deprecated in favour of the new pattern which is what we have. Maybe you're testing this with an older version.

CMake/riscv-esp32c61.json
```diff
@@ -13,35 +13,7 @@
...
-                "API_System.IO.Ports": "ON",
-                "API_System.IO.FileSystem": "ON",
-                "API_nanoFramework.System.IO.Hashing": "ON",
-                "API_nanoFramework.System.Security.Cryptography": "ON",
-                "API_Hardware.Esp32": "ON",
-                "API_nanoFramework.Hardware.Esp32.Rmt": "OFF",
-                "API_nanoFramework.ResourceManager": "ON",
-                "API_nanoFramework.System.Collections": "ON",
-                "API_nanoFramework.System.Text": "ON"
+                }
```
josesimoes: This one still belongs in CMake
```suggestion
                },
                "NF_INTEROP_ASSEMBLIES": null
```

josesimoes (changes requested):

.vscode/tasks.TEMPLATE.json
josesimoes: see here: https://docs.espressif.com/projects/esptool/en/latest/esp32/esptool/advanced-options.html

AdrianSoundy: @josesimoes 

The version of esptool i have is v4.12.dev1
I tried removing all IDF then reinstalling first with the installer EIM( esptool was still the same).
Then with the WIndows Offline installer v5.5.4
I still get version v4.12.dev1 for esptool.

I can't see where it comes from but a constraints file is created in v5.5.4 for tool with the version in it.

So I don't know if that correct or not. I try it on another system.

Anway the 4.12.dev1 release uses underscore for the commands where latest in github uses hyphens.

The funny thing is the 4.12.dev1 version also supports all the new chips.
Can you check your version if you start a IDF 5.5.4 command prompts ?

CMake/riscv-esp32c61.json
AdrianSoundy: Done

.vscode/tasks.TEMPLATE.json
josesimoes: I have v5.1 installed. Here's the print screen. As you can see the pattern for the options is `-` separated...
image

josesimoes: Maybe install from https://github.com/espressif/esptool

AdrianSoundy: I have reverted these changes

PR #3449, opened 2026-06-26
Add PIO support for RP2040 and RP2350

begeistert:
## Description
- Native side of the new nanoFramework.Hardware.Pico class library: gives managed code access to the RP2040/RP2350 PIO blocks.
- Covers program add/remove/claim, state machine init from a config blob, FIFO put/get, FIFO level and PC reads, restart/clkdiv-restart/exec, set clock divisor, and force/clear IRQ.
- PIO state machine interrupts are surfaced to managed code through a NativeEventDispatcher driver, one dispatcher per block.
- Drives the PIO registers straight from CMSIS (no pico-sdk), so the same code serves PIO v0 (RP2040) and v1 (RP2350, with PIO2 and the GPIO base window).
- Hooked into RP_PICO_RP2040, RP_PICO2_RP2350 and RP_PICO_W_RP2040. On the Pico W the IRQ0 -> NVIC wiring is compiled out so the PIO stays reserved for the CYW43.

## Motivation and Context
- The Pico boards had no managed PIO API. This is the firmware half the nanoFramework.Hardware.Pico library binds against.
- Related with nanoframework/Home#1808

## How Has This Been Tested?
- Built RP2040 and RP2350 and ran the PIO samples on the virtual device (FIFO loopback, GPIO out, SM -> managed interrupt).
- On real hardware: blink/fade on a Pico 2 (on-board LED) and on a Pico 1 W (GP0 on a logic analyzer). RP2040 and RP2350 keep building clean.

## Types of changes

## Checklist
- [x] I have read the CONTRIBUTING document.


josesimoes:
@coderabbit review

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native.cpp
```diff
@@ -0,0 +1,202 @@
+#include "nf_hardware_rpi_native.h"
```
josesimoes: Something off with the encoding of this file... please fix.
Missing license header.

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -270,6 +271,17 @@ if(API_Hardware.Stm32)
     PerformSettingsForApiEntry("nanoFramework.Hardware.Stm32")
 endif()
 
+# Hardware.Rp2040
+if(API_Hardware.Rpi)
+    ##### API name here (doted name)
+    PerformSettingsForApiEntry("nanoFramework.Hardware.Rpi")
+
+    # special: the PIO IRQ driver is an internal interrupt driver (no managed end of its own); it is
```
josesimoes:
```suggestion
```

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -270,6 +271,17 @@ if(API_Hardware.Stm32)
...
+    # delivered to managed code through NativeEventDispatcher("PioIrqDriver", ...).
```
josesimoes: No need for any of these comments here as they don't add anything to the build or relevant information.
```suggestion
```

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native.h
```diff
@@ -0,0 +1,275 @@
+//-----------------------------------------------------------------------------
```
josesimoes: Please replace with the standard license header.
Also fix the file encoding.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,219 @@
...
+        hr = CLR_E_INVALID_PARAMETER;
+        return;
+    }
+
+    pio->IRQ_FORCE = (1u << param1);
+}
+
+void PioBlock::NativeClearIrq(signed int param0, signed int param1, HRESULT &hr)
+{
+    PIO_TypeDef *pio = PioFromIndex(param0);
```
josesimoes: This will be distributed as an official lib with support in official firmware.
Please add `<NF_IsCoreLibrary>True</NF_IsCoreLibrary>` to your nfproj so it skips the Interop safety layer.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,401 @@
+//
```
josesimoes, line 1: Please use standard license header

josesimoes (changes requested):
Pretty neat! A few changes required, mostly with license headers and the fact that this is a core library.

begeistert:
Thanks for the review! I really appreciate the feedback! It's pretty late over here, so I'll go through all the comments, apply the requested formatting changes, and push the updates first thing tomorrow morning 😃

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -42,6 +42,7 @@ option(API_nanoFramework.Networking.Thread              "option for nanoFramewor
 
 # Stm32 only
 option(API_Hardware.Stm32                               "option for Hardware.Stm32")
+option(API_Hardware.Rpi                              "option for Hardware.Rpi")
```
josesimoes:
```suggestion
option(API_Hardware.Rpi                                 "option for Hardware.Rpi")
```

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -270,6 +271,14 @@ if(API_Hardware.Stm32)
     PerformSettingsForApiEntry("nanoFramework.Hardware.Stm32")
 endif()
 
+# Hardware.Rpi
+if(API_Hardware.Rpi)
+    PerformSettingsForApiEntry("nanoFramework.Hardware.Rpi")
```
josesimoes: My review on this is wrong, should follow the pattern. I've deleted the comment, but you still picked up... apologies.

```suggestion
    ##### API name here (doted name)
    PerformSettingsForApiEntry("nanoFramework.Hardware.Rpi")
```

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native.cpp
josesimoes, line 1: Looks better now. Still has a bunch of empty lines for no reason...

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico.h
josesimoes, line 1: same here....

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_PioIrqDriver.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // SaveNativeEventToHALQueue touches the shared HAL queue, so take the ISR critical section like GPIO.
+    NATIVE_INTERRUPT_START
+
+    chSysLockFromISR();
+
+    // SM irq flags -> IRQ0_INTS bits [11:8]
+    unsigned int flags = (pio->IRQ0_INTS >> 8) & 0x0Fu;
+    if (flags != 0)
+    {
+        CLR_RT_HeapBlock_NativeEventDispatcher *ctx = s_pioCtx[block];
```
josesimoes: What is the reason you are creating a CLR_RT_HeapBlock_NativeEventDispatcher here? This is not meant to be used by applications... maybe related with the managed code which I haven't seen yet...

begeistert: Spot on regarding the managed code! The native side actually doesn't allocate the dispatcher at all—the managed library handles it lazily.

When an application subscribes to the `PioBlock.Interrupt` event, the managed code creates a `NativeEventDispatcher("PioIrqDriver", block)` to route the state-machine IRQs to the managed handler. This is the exact same pattern used by the GPIO driver.

The `PioIrqDriver` on the native side is just the `CLR_RT_DriverInterruptMethods` that the dispatcher binds to by name. The `CLR_RT_HeapBlock_NativeEventDispatcher *pContext` you see in these procedures is simply the context that the CLR hands down to us, not something we allocate manually. Apps never touch or see the dispatcher directly.

You can check out how the managed side handles this integration here:
[PioBlock.cs#L181-L211](https://github.com/begeistert/nanoFramework.Hardware.Pico/blob/079a53d/nanoFramework.Hardware.Pico/Pio/PioBlock.cs#L181-L211)

That being said, if there is a safer or more idiomatic approach you prefer for handling the dispatcher at the core level, I am completely open to refactoring this part!

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico.h
begeistert: Sorry about that, I was using an script to autogenerate it

CMake/Modules/FindNF_NativeAssemblies.cmake
begeistert: Sorry, I took some comments immediately yesterday before leaving

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native.h
```diff
@@ -0,0 +1,1021 @@
+//
+// Copyright (c) .NET Foundation and Contributors
+// See LICENSE file in the project root for full license information.
+//
+
+
+
+//-----------------------------------------------------------------------------
```
josesimoes: No need for this warning here. Please remove.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico.h
josesimoes, line 1: Stil seeing the double new lines all over the file...
You've mentioned that you're using a script to generate this. No need to. When you build the C# library, these are created for you. Look for the Stubs folder inside bin\Debug (or Release) and copy/paste from there.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    g_PioOutOfReset |= (1u << blockIndex);
+}
+
+HRESULT Library_nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioBlock::
+    NativeAddProgram___STATIC__I4__I4__SZARRAY_U2__I4__I4(CLR_RT_StackFrame &stack)
+{
+    NANOCLR_HEADER();
+    {
```
josesimoes, line 99: You don't need the global {} if you declare all the vars at the beginning of the function. Please do that as it makes the code more readable.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+        PIO_TypeDef *pio = PioFromIndex(block);
+        if (pio == nullptr || sm < 0 || sm > 3)
+        {
+            NANOCLR_SET_AND_LEAVE(CLR_E_INVALID_PARAMETER);
+        }
+        PioEnsureOutOfReset(block);
+
+        // FSTAT TX_FULL = bits [19:16]
+        unsigned int guard = PIO_FIFO_WAIT_LIMIT;
+        while ((pio->FSTAT & (1u << (16 + sm))) && --guard)
```
josesimoes: This loop smells... it can loop forever...
I'm sure there is a better wait to do this.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+        int sm = stack.Arg1().NumericByRef().s4;
+
+        PIO_TypeDef *pio = PioFromIndex(block);
+        if (pio == nullptr || sm < 0 || sm > 3)
+        {
+            NANOCLR_SET_AND_LEAVE(CLR_E_INVALID_PARAMETER);
+        }
+        PioEnsureOutOfReset(block);
+
+        stack.SetResult_Boolean((pio->FSTAT & (1u << (16 + sm))) != 0);
```
josesimoes: Magic number here and the condition it's checking is not obvious on what is doing/checking...

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+        int sm = stack.Arg1().NumericByRef().s4;
+
+        PIO_TypeDef *pio = PioFromIndex(block);
+        if (pio == nullptr || sm < 0 || sm > 3)
+        {
+            NANOCLR_SET_AND_LEAVE(CLR_E_INVALID_PARAMETER);
+        }
+        PioEnsureOutOfReset(block);
+
+        stack.SetResult_Boolean((pio->FSTAT & (1u << (8 + sm))) != 0);
```
josesimoes: Same here

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+    int sm = stack.Arg1().NumericByRef().s4;
+    CLR_RT_HeapBlock_Array *arr = stack.Arg2().DereferenceArray();
+    int offset = stack.Arg3().NumericByRef().s4;
+    int count = stack.Arg4().NumericByRef().s4;
+    int timeoutMs = stack.Arg5().NumericByRef().s4;
+
+    PIO_TypeDef *pio = PioFromIndex(block);
+    PioDmaWork *work = &g_PioDmaWork[block][sm];
+
+    // re-established on each re-entry; SetupTimeoutFromTicks pushes hbTimeout once
```
josesimoes: common pattern in the code base, no need to add any explanation about it.
```suggestion
```

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+    hbTimeout.SetInteger((CLR_INT64)timeoutMs * TIME_CONVERSION__TO_MILLISECONDS);
+    NANOCLR_CHECK_HRESULT(stack.SetupTimeoutFromTicks(hbTimeout, timeoutTicks));
+
+    // first call: validate + arm the DMA + park. jose's pattern: m_customState 1 => setup, 2 => waiting
```
josesimoes:
```suggestion
    // first call: validate + arm the DMA + park
```

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+        ch->channel->TRANS_COUNT = (unsigned int)count;
+        dmaChannelEnableInterruptX(ch);
+        ch->channel->CTRL_TRIG = DMA_CTRL_TRIG_EN | DMA_CTRL_TRIG_DATA_SIZE_WORD | DMA_CTRL_TRIG_INCR_WRITE |
+                                 DMA_CTRL_TRIG_TREQ_SEL(dreq) | DMA_CTRL_TRIG_CHAIN_TO(ch->chnidx);
+
+        work->Channel = ch;
+        work->Buffer = buf;
+        work->Count = (unsigned int)count;
+
+        Events_Get(SYSTEM_EVENT_FLAG_PICOPIO);
```
josesimoes: No need to get this event here... 
Or is there a good reason for calling this API?

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,670 @@
...
+    if (work->Channel != nullptr)
+    {
+        const rp_dma_channel_t *ch = work->Channel;
+
+        // on timeout the channel is still busy; stop it so it can't write after we free the buffer
+        if (dmaChannelIsBusyX(ch))
+        {
+            ch->channel->CTRL_TRIG &= ~DMA_CTRL_TRIG_EN;
+            DMA->CHAN_ABORT = (1u << ch->chnidx);
+            while (DMA->CHAN_ABORT & (1u << ch->chnidx))
```
josesimoes: I really don't like these infinite loops...
You'll tell me that this will always exit because the DMA will always abort the operation. Until the day it doesn't because of a reason and then we'll be stuck here until the watchdog kicks in...
Is this check really needed? or you're just being over thorough?

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_PioIrqDriver.cpp
josesimoes: Still not getting it... Can a C# application subscribe a Pio event? What can it do with it?

Ellerbach:
@begeistert we have changed the directory and target names of the PICO to simplify things.
And also, that's a good news, we just merged the RP2350 Wifi!
So, let me know if you need help to rebase everything properly with the new folder structure. I4ll be happy to assist or do it for you.

begeistert:
Hi @Ellerbach! Apologies for the delay in getting back to you. I went down a rabbit hole optimizing my emulators so I could validate the PIO behavior faster locally. It worked out perfectly in the end, but it took me off the radar for a bit!

(Fun side effect: I actually ended up getting nanoFramework + PIO compiling and running entirely in the browser via WebAssembly to test this! Here is a sneak peek of what I was building while missing running the nanoFramework uf2 with PIO support).

Captura de pantalla 2026-07-23 a la(s) 2 36 58 a m

Great news on the RP2350 WiFi merge! I appreciate the offer to assist with the rebase. I think I did it correctly and pushed the updates, but I really hope I got the new folder structure right! If something looks off or if I broke anything, I will definitely take you up on that offer and ask for your help.

I am back on track now and actively working on @josesimoes's requested changes. Will update the PR in the next few days. Thanks!

begeistert:
Just a quick heads-up: I did a force push to update the branch after the rebase. I really hope this isn't considered a bad practice in this repository!

To be completely honest, I didn't read the contributing guidelines thoroughly beforehand, but I am going to read them right now. Apologies in advance if I messed up the expected workflow!

josesimoes:
No problem at all.
In the mean time know that a lot of work has been happening in ChibiOS regarding the Picos. In particular there was work on the PIO, so it may worth revisiting what's API being offered there. It's the perfect time to provide feedback and introduce changes.

Ellerbach:
@begeistert this looks really cool!

And for the structure, realized today I messed up with one target, the 2040 W, you'll have to rebase and make the adjustements. Sorry for that! And you also have a conflict to solve.

We'll get there! Thanks again for your hard work and contributions!

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico.h
begeistert: Yes, actually I took them from the stubs, hehe maybe I was doing something wrong, but I will be more careful

begeistert:
> @begeistert this looks really cool!
> 
> And for the structure, realized today I messed up with one target, the 2040 W, you'll have to rebase and make the adjustements. Sorry for that! And you also have a conflict to solve.
> 
> We'll get there! Thanks again for your hard work and contributions!

No problem @Ellerbach!

I will rebase the times needed to have the changes in order

And I just want to confirm, I am not absent, I've been testing the changes, and thinking on an answer for @josesimoes questions hehehe

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
begeistert: You are absolutely right, and I completely agree. Raw infinite while loops waiting for hardware registers are a recipe for watchdog resets if the AHB bus or silicon ever glitches.

The reason that check is there is because both datasheets explicitly mandate it. According to Table 149 (RP2040 Datasheet) and Table 1182 (RP2350 Datasheet) under the CHAN_ABORT register description:

"After writing, this register must be polled until it returns all-zero. Until this point, it is unsafe to restart the channel."

If we just abort and immediately proceed to platform_free(work->Buffer) without waiting for those in-flight transfers to flush, the DMA might write into freed memory, causing catastrophic heap corruption.

However, your point stands: we shouldn't trust the hardware infinitely. I will update both NativeRead and NativeWrite to use a guard counter (like PIO_FIFO_WAIT_LIMIT) for the abort loop, so if the DMA controller ever hangs, we gracefully exit instead of locking the thread.

Thanks for catching this! I'll push the fix shortly.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
begeistert: The Events_Get here drains any stale SYSTEM_EVENT_FLAG_PICOPIO before entering the WaitEvents loop — without it, a leftover flag from a previous DMA could cause a spurious wakeup. It follows the same pattern used in the SerialPort drivers (e.g. Events_Get(SYSTEM_EVENT_FLAG_COM_IN) with the comment // clear event by getting it).

Happy to change the approach if there's a preferred alternative though! 😄

CMake/Modules/FindNF_NativeAssemblies.cmake
```diff
@@ -42,6 +42,7 @@ option(API_nanoFramework.Networking.Thread              "option for nanoFramewor
 
 # Stm32 only
 option(API_Hardware.Stm32                               "option for Hardware.Stm32")
+option(API_Hardware.Rpi                              "option for Hardware.Rp2040")
```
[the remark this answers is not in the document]
begeistert: Resolved. The option metadata was updated with the rename; `Hardware.Rp2040` no longer appears in the CMake modules.

Kconfig.apis
```diff
@@ -113,6 +113,11 @@ config API_HARDWARE_STM32
     depends on RTOS_CHIBIOS
     default n
 
+config API_HARDWARE_RPI
+    bool "Hardware.Rp2040"
+    depends on RTOS_CHIBIOS
+    default n
+
```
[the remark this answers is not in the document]
begeistert: Resolved. The prompt now reads `"Hardware.Pico"`, matching `API_HARDWARE_PICO` and the actual hardware scope (RP2040 and RP2350).

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,219 @@
+//
+// Copyright (c) .NET nanoFramework PIO contributors
+//
+// PioBlock InternalCalls. Drives PIO registers directly (CMSIS, no pico-sdk). PIO v0 is a prefix of v1
+// so the same code serves RP2040 and RP2350 (which adds PIO2).
+//
```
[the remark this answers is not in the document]
begeistert: Resolved. The file now starts with the standard `.NET Foundation and Contributors` header.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,219 @@
...
+    if (pio == nullptr || param0 < 0 || param0 > 2)
+    {
+        hr = CLR_E_INVALID_PARAMETER;
+        return -1;
+    }
+
+    // first PIO call in any flow, so un-reset here
+    PioEnsureOutOfReset(param0);
```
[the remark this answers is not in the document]
begeistert: Outdated - this file no longer exists after the rename, and the pattern it flagged is gone with it: the current implementation has no `PioEnsureOutOfReset`; block init and reset release are owned by the ChibiOS RP PIO LLD (driver init / `pioSmAlloc`), independent of any managed call ordering.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    unsigned short *instr = param1.GetBuffer();
+    int length = param2;
+    int origin = param3;
+
+    if (length <= 0 || length > 32)
+    {
+        return -1;
+    }
```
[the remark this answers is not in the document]
begeistert: Outdated file, but the substance is addressed in its successor: `AddProgram` validates the array reference, `length <= 0 || length > 32`, coherence with `m_numOfElements`, and checks the loader result before returning an offset.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioBlock.h
```diff
@@ -0,0 +1,41 @@
...
+//                   ** WARNING! ** 
+//    This file was generated automatically by a tool.
+//    Re-running the tool will overwrite this file.
+//    You should copy this file to a custom location
+//    before adding any customization in the copy to
+//    prevent loss of your changes when the tool is
+//    re-run.
+//
+//-----------------------------------------------------------------------------
+
```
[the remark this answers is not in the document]
begeistert: Outdated - the generated Rpi header no longer exists; the current generated sources carry the standard repository header.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,401 @@
...
+        (b[PIO_CFG_AUTOPUSH] << 16) | (b[PIO_CFG_AUTOPULL] << 17) | (pushThresh << 20) |
+        (pullThresh << 25) | ((join & 3u) << 30) | ((join >> 2) << 14);
+
+#if defined(RP2350)
+    // GPIOBASE is 0 or 16; CMSIS types it __I so write through a volatile pointer
+    *(volatile unsigned int *)&pio->GPIOBASE = b[PIO_CFG_GPIO_BASE];
+#endif
+
+    // set PC via JMP <offset> (opcode 0)
+    pio->SM[sm].INSTR = (unsigned int)(param2 & 0x1F);
```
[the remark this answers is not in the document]
begeistert: By design, under the project's validation model: every blob field originates from managed config setters that are range-checked at construction time (counts, thresholds, wrap, pin windows via `CheckWindow`), and the native side masks each field to its hardware-legal range (`& 0x3F`, `& 0x7`, `& 0x1F`, `PioRelPin`) as a second line of defense. The only way to deliver an out-of-range field is forging the blob via reflection, and even then the masking guarantees a defined register state - garbage in, defined garbage out, never corruption. If strict rejection is preferred it is a handful of comparisons on a cold path (`NativeInit` runs once per configuration) - happy to add it if that is the maintainers' preference.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,401 @@
...
+        // PINCTRL: SET_COUNT [28:26], SET_BASE [9:5]
+        pio->SM[sm].PINCTRL = ((unsigned int)chunk << 26) | ((unsigned int)pin << 5);
+        // SET pindirs, dirs
+        pio->SM[sm].INSTR = 0xE000u | (4u << 5) | (dirs & 0x1Fu);
+
+        remaining -= chunk;
+        pin += chunk;
+    }
+
+    pio->SM[sm].PINCTRL = savedPinCtrl;
```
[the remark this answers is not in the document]
begeistert: Resolved in the successor implementation: `SetConsecutivePinDirs` validates `basePin`, `count`, and the full GPIOBASE window (`basePin < gpioBase || basePin - gpioBase + count > 32`) before executing anything. This path is exercised on real hardware by our API battery, including the RP2350 `GPIOBASE = 16` window.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native_nanoFramework_Hardware_Rpi_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,401 @@
...
+    PIO_TypeDef *pio = PioFromIndex(param0);
+    int sm = param1;
+    if (pio == nullptr || sm < 0 || sm > 3)
+    {
+        hr = CLR_E_INVALID_PARAMETER;
+        return;
+    }
+
+    pio->SM[sm].CLKDIV = ((unsigned int)param2 << 16) | ((unsigned int)param3 << 8);
+    pio->CTRL |= (1u << (8 + sm));
```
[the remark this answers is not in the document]
begeistert: Resolved in the successor implementation: the property setter validates 1.0..65536.0 natively and rejects out-of-range values with `CLR_E_OUT_OF_RANGE`; the blob path receives values already validated by the managed config and packs them masked to the INT/FRAC fields.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_PioIrqDriver.cpp
```diff
@@ -0,0 +1,164 @@
...
+static HRESULT PioIrqInitialize(CLR_RT_HeapBlock_NativeEventDispatcher *pContext, unsigned __int64 userData)
+{
+    int block = (int)(userData & 0xFF);
+    if (block < 0 || block > 2 || PioFromIndex(block) == nullptr)
+    {
+        return CLR_E_INVALID_PARAMETER;
+    }
+
+    s_pioCtx[block] = pContext;
+    return S_OK;
```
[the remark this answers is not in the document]
begeistert: Outdated - `PioIrqDriver.cpp` no longer exists. Interrupt dispatch now uses the ChibiOS per-state-machine callbacks registered at `pioSmAlloc`, with a single idempotent block handler that acknowledges latched and forced flags separately; there is no dispatcher registry to overwrite.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_PioIrqDriver.cpp
```diff
@@ -0,0 +1,164 @@
...
+    }
+    else
+    {
+        pio->IRQ0_INTE &= ~(0x0Fu << 8);
+        nvicDisableVector(vector);
+    }
+#else
+    (void)fEnable;
+    (void)pio;
+#endif
```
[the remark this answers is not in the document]
begeistert: Outdated - same file removed; no conditional stub path remains in the current dispatch design.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_target.h
```diff
@@ -0,0 +1,30 @@
+//
+// Copyright (c) .NET nanoFramework PIO contributors
+//
+// Shared helpers for the RP2040/RP2350 PIO native interop. Include after the CMSIS header
+// (rp2040.h / rp2350.h) so PIO_TypeDef and PIO0/PIO1/PIO2 are defined.
+//
```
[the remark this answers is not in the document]
begeistert: Resolved. The file starts with the standard `.NET Foundation and Contributors` header.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Rpi/nf_hardware_rpi_native.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+// clang-format off
```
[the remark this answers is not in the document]
begeistert: Outdated - the file no longer exists; its successor carries the standard header.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,523 @@
...
+        bool output = stack.Arg4().NumericByRef().u1;
+
+        PIO_TypeDef *pio = PioFromIndex(block);
+#if defined(RP2350)
+        const int maxPins = 48;
+#else
+        const int maxPins = 30;
+#endif
+        if (pio == nullptr || sm < 0 || sm > 3 || basePin < 0 || count < 0 || basePin > maxPins ||
+            count > maxPins - basePin)
```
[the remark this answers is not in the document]
begeistert: Resolved. `SetConsecutivePinDirs` now validates the pin against the state machine's GPIOBASE window before anything executes, and the LLD translates through `pioGpioToRel`, so an absolute GPIO can no longer alias modulo 32. Verified on hardware with `GPIOBASE = 16` (driven-pin jmp tests on GP16 in our API battery).

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
begeistert: Closing the loop on the rationale above: this `Events_Get` only drains a stale `SYSTEM_EVENT_FLAG_PICOPIO` left over from a previous transfer on another state machine or channel, so the first `WaitEvents` cannot wake spuriously and burn a re-entry for nothing. It is a one-flag read on the cold path (transfer arming), not a wait. This shape has been exercised heavily on real hardware (full API battery on both chips plus long soak runs) without a miswake. Happy to drop it if you prefer strict symmetry.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_native_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
begeistert: Agreed, and fixed: there are no raw register spins left in the binding. The transfer wait is now the `WaitEvents` pattern bounded by the caller's timeout (`eventResult == false` breaks out), and the TX-drain helper is bounded by an iteration limit that returns `CLR_E_TIMEOUT`. Every wait either wakes on the interrupt/event or exits by timeout - nothing can sit there until the watchdog.

emolitor:
@begeistert I'm also looking at adding a better interface for PIO IRQ flags which I think will simplify your code as well as adding the missing FIFO modes. I will tag you in an MR shortly to see if you think this will help.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,309 @@
...
+    CLR_RT_HeapBlock *pThis = stack.This();
+    FAULT_ON_NULL(pThis);
+
+    block = pThis[FIELD___index].NumericByRef().s4;
+
+    VALIDATE_PIO_BLOCK(block);
+
+    program_array =
+        pProgram
+            [Library_nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioProgram::FIELD__Instructions]
```
josesimoes, line 75: Better use a typedef with the struct name to shorten it (like PioProgram) to improve readability. (other events of this in the code base)

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioBlock.cpp
```diff
@@ -0,0 +1,309 @@
...
+}
+
+HRESULT Library_nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioBlock::
+    RemoveProgram___VOID__nanoFrameworkHardwarePicoPioPioProgram__U4(CLR_RT_StackFrame &stack)
+{
+    NANOCLR_HEADER();
+
+    uint8_t block;
+    int length;
+    CLR_RT_HeapBlock_Array *program;
```
josesimoes, line 116: Can you please improve program and pProgram var names? They don't convey the meaning of the var content and also can be easily confused...

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,830 @@
...
+    }
+    else
+    {
+        pioSmDisableX(ctx.smp);
+    }
+
+    NANOCLR_NOCLEANUP();
+}
+
+static inline bool PioFifoBlocked(const rp_pio_sm_t *smp, const PioFifoDir dir)
```
josesimoes, line 425: Can you please move the helper functions to the top of the file? Or at the bottom of the file and declare them as static in the Library_nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioStateMachine  struct (following the  //--// there)

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
```diff
@@ -0,0 +1,830 @@
...
+}
+
+static HRESULT PioFifoTransfer(CLR_RT_StackFrame &stack, const PioFifoDir dir)
+{
+    NANOCLR_HEADER();
+
+    PioSmContext ctx{};
+    CLR_RT_HeapBlock hbTimeout{};
+    CLR_INT64 *timeoutTicks;
+    const rp_pio_block_t *blockp = nullptr;
```
josesimoes, line 437: Var name not explicit nor following code style. Maybe pioBlock? 
No need to use cryptic or super abbreviated names for vars. Better use clear and descriptive names. Variables name length is free 😉

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_target.h
```diff
@@ -0,0 +1,231 @@
...
+void PioDmaReleaseSm(int block, int sm);
+
+struct PioSmContext
+{
+    const rp_pio_sm_t *smp;
+    int block;
+    int sm;
+};
+
+inline HRESULT PioGetSmContext(CLR_RT_HeapBlock *pThis, PioSmContext &ctx, const bool allowDisposed = false)
```
josesimoes, line 65: This is a large code block. Doesn't seem OK to declare it inline. Better move it to the cpp and let the compiler figure it it out.

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_target.h
```diff
@@ -0,0 +1,231 @@
...
+    PADS_BANK0->GPIO[gpio] = padbits | RP_PIO_PAD_ISO;
+    IO_BANK0->GPIO[gpio].CTRL = funcsel;
+    PADS_BANK0->GPIO[gpio] = padbits;
+#else
+    PADS_BANK0->GPIO[gpio] = padbits;
+    IO_BANK0->GPIO[gpio].CTRL = funcsel;
+#endif
+}
+
+inline bool NfPioSmDrainTxFifo(const rp_pio_sm_t *smp, unsigned int spinLimit)
```
josesimoes, line 194: Same here. And this one is only called once.

josesimoes (commented):
Getting closer! A few more comments to address. Nothing major.
(Apologies for the delay, but I'm currently away)

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_target.h
begeistert: This one has been deleted thanks to the new helpers from chibiOS 😄

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioBlock.cpp
begeistert: Done

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioBlock.cpp
begeistert: Sure, I was thinking about using `program` for the heap block and `instructions` for the array (following the names in the managed code)

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_nanoFramework_Hardware_Pico_Pio_PioStateMachine.cpp
begeistert: I moved them to the bottom, but this specific helper uses an enum defined inside the cpp file, do you have an example about how this kind of helpers are declared into the structs?

targets/ChibiOS/_nanoCLR/nanoFramework.Hardware.Pico/nanoFramework_hardware_pico_target.h
begeistert: Agreed, let me move it

Banjobeni:
As a PIO newbie, when I read the blinky sample, my interpretation is that `OutShift()` issues an instruction, because it's called on `PioAssembler` (just like `Out()` or `Pull()` is).

```c#
            var asm = new PioAssembler();
            PioLabel loop = asm.DefineLabel();
            asm.MarkLabel(loop);
            asm.Pull(ifEmpty: false, block: true);
            asm.Out(PioDest.Pins, 1);
            asm.Jmp(PioCondition.Always, loop);
            asm.OutShift(PioShiftDir.Right, autoPull: false, threshold: 32);
            PioProgram program = asm.Build();
```

I'd suggest to rename it to `SetOutShift()` and move it to `PioProgram` or `PioStateMachine`, as it seems to be more of a configuration thing rather than an assembler thing.

PR #3452, opened 2026-06-27
Fix closed generic instance corruption and generic type-param cast

danielmeza:
## Description

- Guards the object-lock cache against closed generic instances so that entering a `Synchronized` instance method no longer overwrites the closed `TypeSpec` stored in the object header.
- Completes the `DATATYPE_GENERICINST` handling for the dedicated heap representation (boxed `Nullable<T>`): garbage-collector reachability marking and heap-compaction relocation now walk its fields, and type recovery (`InitializeFromObject`/`ExtractTypeIndexFromObject`), `initobj`/clone, `GetHashCode`, `ObjectsEqual` and `Compare_Values` all treat it as an inline value type.
- Recovers the closed `TypeSpec` during `CEE_CALL`/`CEE_CALLVIRT` so a generic type-parameter cast (`castclass !0`) resolves correctly when the interface call is dispatched within the same assembly.
- Removes the unconditional `NewGenericInstanceObject` debug traces.

## Motivation and Context

- This addresses two crashes that affect closed generic class instances under the generics public preview.
- A closed generic class instance stores its closed `TypeSpec` in the object-header slot that overlaps the monitor-lock pointer. Entering a `Synchronized` instance method — which is what a C# field-like event `add`/`remove` accessor compiles to — locked `this` and wrote the lock pointer into that slot, destroying the `TypeSpec`. Any subsequent generic field or method resolution then dereferenced a wild pointer and faulted (`LoadProhibited`). `FindLockObject` had the inverse defect, reading the `TypeSpec` back as a lock pointer. Monitor ownership still works through the per-thread lock lists, so guarding the cache against generic instances is sufficient.
- A `(T)item` cast (`castclass !0`) inside an explicit interface implementation on a closed generic instance threw `CLR_E_INVALID_CAST`: the dispatch lost the closed `TypeSpec`, so the `VAR` branch of `InitializeFromSignatureToken` had no generic context. The `CALLVIRT` generic-context recovery pre-pass was guarded to `TBL_MethodRef` (cross-assembly interface calls); when the interface and the caller live in the same assembly the call is a `TBL_MethodDef`, so the pre-pass never ran and the generic type stayed null. The fix prefers the receiver's own stored closed `TypeSpec` over the caller-assembly `TypeSpec`-table scan, and adds an else-branch recovery from the receiver for the same-assembly `MethodDef` interface dispatch the pre-pass misses.

## How Has This Been Tested?

- Validated on an ESP32-S3 device build.
- A C# field-like event on a closed generic type (`Observable<int>.Changed += handler; raise`) now subscribes and fires without faulting, where it previously crashed with `LoadProhibited`.
- A generic `DataTemplate<T>` provider whose `IDataTemplate.Build(object)` performs a `(T)item` cast, driving a repeater over an `ObservableCollection<T>` (both in the same assembly), now builds, rebuilds on collection change, and applies per-item bindings with no `InvalidCastException`.

## Types of changes

## Checklist:


src/CLR/Core/Execution.cpp
```diff
@@ -2822,12 +2830,6 @@ HRESULT CLR_RT_ExecutionEngine::NewGenericInstanceObject(
 
     fieldCursor = reinterpret_cast<CLR_RT_HeapBlock *>(giHeader) + totFields;
 
-    CLR_Debug::Printf(
```
josesimoes: These debug snippets are to be kept.

josesimoes (commented):
Apart from this, please review the comments added. They are VERY verbose. I understand AI is eager to add them, but it doesn't make sense otherwise we'll end up with more comments than code making it unreadable. Keep only the absolutely necessary ones and make them as concise as possible.

josesimoes:
@danielmeza please instruct your agent to respect the PR template. Do not remove the checklist items not checked. 😉

josesimoes:
@danielmeza we need a unit test (or more) to cover this. Please add it to mscorlib solution (develop branch). Maybe in the Types project...
The PR adding those will fail (expected) and then this will fix it.

PR #3524, opened 2026-09-06
Report program state in RTM builds

### LINKED ISSUE #1842: RTM builds never report ProgramRunning or ProgramExited, so clients see the device as stuck in Initialize
**Describe the bug**

An RTM build never sets the `StateProgramRunning` or `StateProgramExited` bits of the debugger state mask, because both are guarded by `NANOCLR_ENABLE_SOURCELEVELDEBUGGING`:

- `src/CLR/Core/Execution.cpp`, in `CLR_RT_ExecutionEngine::Execute()`
- `src/CLR/Startup/CLRStartup.cpp`, in `ClrStartup()`

On the client side `nanoFramework.Tools.Debugger.WireProtocol.Commands.DebuggingExecutionChangeConditions.State.Initialize` is `0x00000000`, so a state mask that never gains `ProgramRunning` (`0x00000400`) is indistinguishable from a device still initializing. The device looks stuck before its program starts, for the whole time the program is in fact running.

**Observed effect**

`nanoff --filedeployment` against an RTM build fails with `E2002` and reboots the target repeatedly instead of deploying. A debug build of the same firmware on the same hardware deploys normally.

**Expected behaviour**

The state mask reports the program state regardless of whether source level debugging is compiled in. The mask is state, not a debugging feature — an RTM device is still expected to answer state queries over the Wire Protocol, which is how the deployment tooling decides that the device is ready.

**Proposed fix**

Set `StateProgramRunning` and `StateProgramExited` unconditionally, and leave the `CLR_DBG_Commands_c_Monitor_ProgramExit` broadcast behind the source level debugging guard, since that one is an event rather than state. PR follows.

dkovyazin:
## Description

- `StateProgramRunning` and `StateProgramExited` are now set in every build, not only when `NANOCLR_ENABLE_SOURCELEVELDEBUGGING` is defined.
- The `CLR_DBG_Commands_c_Monitor_ProgramExit` broadcast stays behind that guard, since it is a debugging event rather than state.

## Motivation and Context

Both bits of the debugger state mask were guarded by `NANOCLR_ENABLE_SOURCELEVELDEBUGGING`, so an RTM build never reported either of them. On the client side `State.Initialize` is `0x00000000`, which makes a mask that never gains `ProgramRunning` (`0x00000400`) indistinguishable from a device that is still initializing. An RTM device therefore looks stuck before its program starts, for the entire time the program is actually running.

The visible effect is that `nanoff --filedeployment` against an RTM build fails with `E2002` and reboots the target repeatedly, while a debug build of the same firmware on the same hardware deploys normally.

- Fixes nanoFramework/Home#1842

## How Has This Been Tested?

- Reproduced on an ESP32-S3 target: with an RTM build, deploying files over the Wire Protocol did not go through, while a debug build of the same firmware on the same hardware deployed normally.
- Firmware carrying this change has been in use on that hardware since, with file deployment working on RTM builds.
- Compile-checked against current `main` for `ESP32_S3_QUAD`.

## Types of changes

## Checklist

### DIFF
--- a/src/CLR/Core/Execution.cpp
+++ b/src/CLR/Core/Execution.cpp
@@ -638,10 +638,6 @@ HRESULT CLR_RT_ExecutionEngine::Execute(wchar_t *entryPointArgs, int maxContextS
 
     NANOCLR_CHECK_HRESULT(WaitForDebugger());
 
-    // set unconditionally (not only with source-level debugging): the state mask
-    // zero-value means "initialize", so an RTM build that never sets ProgramRunning
-    // is treated as stuck-in-initialize by debugger clients (nanoff refuses
-    // --filedeployment with E2002 and keeps rebooting the device)
     CLR_EE_DBG_SET_MASK(StateProgramRunning, StateMask);
 
     NANOCLR_CHECK_HRESULT(CLR_RT_HeapBlock_Delegate::CreateInstance(ref, g_CLR_RT_TypeSystem.m_entryPoint, NULL));
--- a/src/CLR/Core/Execution.cpp
+++ b/src/CLR/Core/Execution.cpp
@@ -638,9 +638,11 @@ HRESULT CLR_RT_ExecutionEngine::Execute(wchar_t *entryPointArgs, int maxContextS
 
     NANOCLR_CHECK_HRESULT(WaitForDebugger());
 
-#if defined(NANOCLR_ENABLE_SOURCELEVELDEBUGGING)
+    // set unconditionally (not only with source-level debugging): the state mask
+    // zero-value means "initialize", so an RTM build that never sets ProgramRunning
+    // is treated as stuck-in-initialize by debugger clients (nanoff refuses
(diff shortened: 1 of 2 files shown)


josesimoes (changes requested):
The fix looks OK. Just remove the verbose comment. The reasoning is perfectly explained in the PR comments where it belongs.

dkovyazin:
Comments removed in 8b89a2a4 — the reasoning stays in the PR description.

josesimoes (approved):
LGTM

dkovyazin:
The two red jobs (`Build_ESP32_targets ESP32_C5_THREAD` and `ESP32_C6_THREAD`, build 65638) failed in `Install IDF`, not on the code: the RISC-V toolchain download returned `HTTP Error 504: Gateway Time-out` from the GitHub release and ran out of retries. Everything else in that run is green, including `Check_Code_Style` and the STM32 nightly targets. A rerun should clear it.

PR #3525, opened 2026-09-06
Size the USB-JTAG RX buffer to hold a full Wire Protocol frame

### LINKED ISSUE #1843: USB-Serial-JTAG RX buffer is smaller than a Wire Protocol frame, so bytes are dropped during file deployment
**Describe the bug**

In `targets/ESP32/_common/WireProtocol_HAL_Interface.c` the USB-Serial-JTAG driver is installed with

```c
#define USB_JTAG_BUFFER_SIZE 256
```

A full Wire Protocol frame is 1056 bytes — a 32 byte header plus `WP_PACKET_SIZE` (1024) of payload — so a single frame does not fit in the RX ring. The driver ISR discards incoming bytes once the ring is full, which is exactly what happens while the receiver task is still busy with the previous message rather than draining the ring.

**Observed effect**

Deploying files over the Wire Protocol to an ESP32-S3 connected over USB-Serial-JTAG stalls or fails once a file is large enough to be split into several packets, because storage writes for one packet keep the receiver task busy while the next one arrives. Smaller transfers, where the task is idle between frames, are unaffected.

**Expected behaviour**

The RX ring holds at least one complete frame, so a frame that arrives while the previous one is being processed is buffered rather than truncated.

**Proposed fix**

Raise `USB_JTAG_BUFFER_SIZE` to 2048, which covers a full frame with margin. PR follows.

dkovyazin:
## Description

- `USB_JTAG_BUFFER_SIZE` raised from 256 to 2048 bytes, so the RX ring holds a complete Wire Protocol frame.

## Motivation and Context

A full frame is 1056 bytes — a 32 byte header plus `WP_PACKET_SIZE` (1024) of payload — and did not fit in the 256 byte ring. The USB-Serial-JTAG driver ISR drops incoming bytes once the ring is full, which is what happens while the receiver task is busy with the previous message instead of draining the ring.

The result is that deploying files over the Wire Protocol to a target connected over USB-Serial-JTAG stalls or fails as soon as a file is large enough to be split into several packets, since the storage write for one packet overlaps the arrival of the next. Transfers where the task is idle between frames are unaffected, which is why this only shows up on larger files.

- Fixes nanoFramework/Home#1843

## How Has This Been Tested?

- Reproduced on an ESP32-S3 connected over USB-Serial-JTAG: deploying files of a few tens of kilobytes over the Wire Protocol stalled, while smaller transfers went through.
- Firmware carrying this change has been in use on that hardware since, and those deployments complete.
- Compile-checked against current `main` for `ESP32_S3_QUAD`.

## Types of changes

## Checklist

### DIFF
--- a/targets/ESP32/_common/WireProtocol_HAL_Interface.c
+++ b/targets/ESP32/_common/WireProtocol_HAL_Interface.c
@@ -161,11 +161,7 @@ static uint8_t WP_TransmitMessageTinyUsb(WP_Message *message)
 
 #include "driver/usb_serial_jtag.h"
 
-// must hold at least one full WP frame (32-byte header + WP_PACKET_SIZE payload
-// = 1056 bytes): the driver ISR drops RX bytes when the ring buffer is full,
-// which happens while the receiver task is busy processing the previous message
-// (e.g. Monitor_StorageOperation writing to SD during nanoff --filedeployment)
-#define USB_JTAG_BUFFER_SIZE 2048
+#define USB_JTAG_BUFFER_SIZE (sizeof(WP_Packet) + WP_PACKET_SIZE)
 
 static size_t UsbSerialWrite(const uint8_t *data, size_t dataSize, TickType_t xTicksToWait);
 static size_t UsbSerialRead(uint8_t *data, size_t dataSize, TickType_t xTicksToWait);
--- a/targets/ESP32/_common/WireProtocol_HAL_Interface.c
+++ b/targets/ESP32/_common/WireProtocol_HAL_Interface.c
@@ -161,7 +161,11 @@ static uint8_t WP_TransmitMessageTinyUsb(WP_Message *message)
 
 #include "driver/usb_serial_jtag.h"
 
-#define USB_JTAG_BUFFER_SIZE 256
+// must hold at least one full WP frame (32-byte header + WP_PACKET_SIZE payload
+// = 1056 bytes): the driver ISR drops RX bytes when the ring buffer is full,
+// which happens while the receiver task is busy processing the previous message
+// (e.g. Monitor_StorageOperation writing to SD during nanoff --filedeployment)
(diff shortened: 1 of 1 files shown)


dkovyazin:
One clarification on the description, since the constant is shared: `USB_JTAG_BUFFER_SIZE` is passed as both `rx_buffer_size` and `tx_buffer_size` to `usb_serial_jtag_driver_install()`, and it also caps a single write in `UsbSerialWrite()`. So raising it to 2048 does three things, not one:

- the RX ring finally holds a complete 1056 byte frame, which is the fix;
- the TX ring grows the same way, which costs memory but is otherwise harmless;
- writes are chunked at 2048 instead of 256, so a reply goes out in fewer, larger pieces.

The memory cost is 1792 bytes per direction, 3584 bytes of DRAM in total. If that is unwelcome, splitting the constant into separate RX and TX sizes and raising only the RX one would keep the fix at half the cost — happy to change it if you prefer that shape.

josesimoes (changes requested):
I understand the fix, it makese sense.
Let's not use magic numbers please. As WP packet lenght can be adjusted by a configuration, please use that one here instead of a magic number. Better add the JTAG packet with the WP lenght and have the compiler do tha math.

Please remove the comment, the reasoning is already documented in the PR comments, where it belongs.

dkovyazin:
Done in 87044135. The buffer is now `sizeof(WP_Packet) + WP_PACKET_SIZE`, the same expression `receptionBuffer` in WireProtocol_App_Interface.c is built from, so it follows a target that overrides `WP_PACKET_SIZE` in target_common.h. With the ESP32 default of 1024 that is 1056 bytes instead of the 2048 I had hardcoded — one full frame, and still well above the 64 byte minimum the USB Serial JTAG driver enforces. Comment removed.

josesimoes (approved):
LGTM

PR #3528, opened 2026-09-06
Transfer external RAM buffers through an internal RAM bounce buffer

### LINKED ISSUE #1846: ESP32: SD file IO from managed code is capped at ~150-250 KB/s because the buffers live in PSRAM
**Describe the problem**

On targets with external RAM, file IO through the SD card is capped at roughly 150-250 KB/s regardless of the card or the bus width.

The SDMMC host cannot DMA to or from external RAM. When `sdmmc_cmd` is handed a buffer that lives in PSRAM it falls back to transferring one sector at a time through a small internal temporary buffer, which costs around 2 ms per 512 byte sector. The CLR managed heap lives in PSRAM, so every buffer that arrives from managed code takes that path: `LITTLEFS_FS_Driver::Read()` and `Write()` pass the caller's pointer straight to `fread`/`fwrite`.

**Expected behaviour**

Large transfers reach the throughput the card and the bus are capable of, rather than being limited by the sector-at-a-time fallback.

**Proposed fix**

Copy through a bounce buffer allocated in internal DMA-capable RAM when the caller's buffer is in external RAM and the transfer is large enough to matter, which restores multi sector DMA. Small transfers (under one sector) keep going straight through, since the stdio buffer serves them without touching the medium. The extra memcpy is negligible next to the bus transfer, and file system driver calls are serialised by the CLR, so one shared buffer is enough. PR follows.

dkovyazin:
## Description

- Transfers in `Read()` and `Write()` go through a bounce buffer in internal DMA-capable RAM when the caller's buffer lives in external RAM and the transfer is at least one sector.
- Smaller transfers and buffers in internal RAM keep going straight through, unchanged.

## Motivation and Context

The SDMMC host cannot DMA to or from external RAM. Handed a buffer in PSRAM, `sdmmc_cmd` falls back to transferring one sector at a time through a small internal temporary buffer, at roughly 2 ms per 512 byte sector, which caps file IO at about 150-250 KB/s. The CLR managed heap lives in PSRAM, so every buffer arriving from managed code takes that path, since the driver passes the caller's pointer straight to `fread`/`fwrite`.

Copying through an internal RAM buffer restores multi sector DMA. The extra memcpy is negligible next to the bus transfer, and file system driver calls are serialised by the CLR, so a single shared buffer is enough. It is allocated lazily and, if internal RAM has run out, the caller simply falls back to the direct path.

- Fixes nanoFramework/Home#1846

## How Has This Been Tested?

- Built for an ESP32-S3 target against current `main`, including the link step.
- Exercised on hardware with an ESP32-S3 and PSRAM, reading and writing files on an SD card from managed code.

## Types of changes

## Checklist

### DIFF
--- a/targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
+++ b/targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
@@ -347,7 +347,15 @@ HRESULT LITTLEFS_FS_Driver::Read(void *handle, uint8_t *buffer, int size, int *b
 
         while (total < size)
         {
-            int pad = (int)(ftell(fileHandle->file) & 3);
+            long filePosition = ftell(fileHandle->file);
+
+            if (filePosition < 0)
+            {
+                NANOCLR_SET_AND_LEAVE(CLR_E_FILE_IO);
+            }
+
+            // keeps the pointer handed to the whole sector transfers 4 byte aligned
+            int pad = (int)(filePosition & 3);
 
             int chunk = size - total;
             if (chunk > c_ioBounceBufferSize - pad)
@@ -432,7 +440,15 @@ HRESULT LITTLEFS_FS_Driver::Write(void *handle, uint8_t *buffer, int size, int *
 
         while (total < size)
         {
-            int pad = (int)(ftell(fileHandle->file) & 3);
+            long filePosition = ftell(fileHandle->file);
+
+            if (filePosition < 0)
+            {
+                NANOCLR_SET_AND_LEAVE(CLR_E_FILE_IO);
+            }
+
+            // keeps the pointer handed to the whole sector transfers 4 byte aligned
+            int pad = (int)(filePosition & 3);
 
             int chunk = size - total;
             if (chunk > c_ioBounceBufferSize - pad)
--- a/targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
+++ b/targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
@@ -8,32 +8,58 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
+#include <sdkconfig.h>
+
+#if CONFIG_SPIRAM
 #include <esp_heap_caps.h>
 #include <esp_memory_utils.h>
+#endif
 
 extern FileSystemVolume *g_FS_Volumes;
 
 static int32_t RemoveAllFiles(const char *path);
 static int NormalizePath(const char *root, const char *path, char *buffer, size_t bufferSize);
 
-// bounce buffer in internal RAM, for transfers of buffers backed by external RAM
-static uint8_t *s_ioBounceBuffer = NULL;
 static const int c_ioBounceBufferSize = 16 * 1024;
+
+#if CONFIG_SPIRAM
 
 // transfers below one sector are left to the stdio buffer, the bounce path would only add an ftell
 static const int c_ioBounceMinTransfer = 512;
 
-static uint8_t *GetIoBounceBuffer()
+// bounce buffer in internal RAM, for transfers of buffers backed by external RAM
+static uint8_t *s_ioBounceBuffer = NULL;
+
+// NULL when the transfer needs no bouncing, or when there is no internal RAM left: the caller then
+// takes the direct path
+static uint8_t *GetIoBounceBuffer(const void *buffer, int size)
 {
+    if (size < c_ioBounceMinTransfer || !esp_ptr_external_ram(buffer))
+    {
+        return NULL;
+    }
+
     if (s_ioBounceBuffer == NULL)
     {
         s_ioBounceBuffer =
             (uint8_t *)heap_caps_malloc(c_ioBounceBufferSize, MALLOC_CAP_INTERNAL | MALLOC_CAP_DMA | MALLOC_CAP_8BIT);
     }
 
-    // NULL, when there is no internal RAM left, makes the caller fall back to the direct path
     return s_ioBounceBuffer;
 }
+
+#else
+
+// no external RAM on this target, a transfer never needs bouncing
+static uint8_t *GetIoBounceBuffer(const void *buffer, int size)
+{
+    (void)buffer;
+    (void)size;
+
+    return NULL;
+}
+
+#endif
 
 bool LITTLEFS_FS_Driver::LoadMedia(const void *driverInterface)
 {
@@ -306,10 +332,7 @@ HRESULT LITTLEFS_FS_Driver::Read(void *handle, uint8_t *buffer, int size, int *b
 
     fileHandle = (LITTLEFS_FileHandle *)handle;
 
-    if (size >= c_ioBounceMinTransfer && esp_ptr_external_ram(buffer))
-    {
-        readBounce = GetIoBounceBuffer();
-    }
+    readBounce = GetIoBounceBuffer(buffer, size);
 
     if (readBounce != NULL)
     {
@@ -387,10 +410,7 @@ HRESULT LITTLEFS_FS_Driver::Write(void *handle, uint8_t *buffer, int size, int *
 
     fileHandle = (LITTLEFS_FileHandle *)handle;
 
-    if (size >= c_ioBounceMinTransfer && esp_ptr_external_ram(buffer))
-    {
-        writeBounce = GetIoBounceBuffer();
-    }
+    writeBounce = GetIoBounceBuffer(buffer, size);
 
     if (writeBounce != NULL)
     {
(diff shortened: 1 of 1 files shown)


targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
```diff
@@ -282,8 +319,51 @@ HRESULT LITTLEFS_FS_Driver::Read(void *handle, uint8_t *buffer, int size, int *b
...
     fileHandle = (LITTLEFS_FileHandle *)handle;
 
-    // read from the file
-    readCount = fread(buffer, 1, size, fileHandle->file);
+    // a buffer in external RAM is read through the internal RAM bounce buffer
+    // (see above). pad = (position & 3) keeps the pointer used for direct whole
+    // sector transfers 4 byte aligned. Small reads (< 512) go straight through:
+    // the stdio buffer serves them without touching the medium, while the bounce
+    // path would add an ftell on every call.
+    if (size >= 512 && esp_ptr_external_ram(buffer))
```
josesimoes: Do we have a DEFINE for this 512? Better replace with that so it gets self explained.

targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
```diff
@@ -324,10 +405,47 @@ HRESULT LITTLEFS_FS_Driver::Write(void *handle, uint8_t *buffer, int size, int *
 
     fileHandle = (LITTLEFS_FileHandle *)handle;
 
-    // write to the file
-    writeCount = fwrite(buffer, 1, size, fileHandle->file);
+    // a buffer in external RAM is written through the internal RAM bounce
+    // buffer, for the same reason as in Read
+    if (size >= 512 && esp_ptr_external_ram(buffer))
```
josesimoes: Same here: do we have a DEFINE for this 512? Better replace with that so it gets self explained.

josesimoes (changes requested):
Please remove the verbose comments. The reasoning should be explained in the PR comments. If needed/missing, move it from the comments in the source files to the PR comments.

josesimoes:
@coderabbitai review

targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
dkovyazin: There was no define for it, so 59eb8396 adds one: `c_ioBounceMinTransfer` (512, one sector). Below that the stdio buffer serves the call without touching the medium, while the bounce path would add an `ftell` to every read.

targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
dkovyazin: Same constant here, `c_ioBounceMinTransfer` in 59eb8396.

targets/ESP32/_littlefs/littlefs_FS_Driver.cpp
```diff
@@ -6,12 +6,41 @@
...
+// The SDMMC host cannot DMA from or to external RAM: for a PSRAM buffer,
+// sdmmc_cmd falls back to single sector transfers through a 512 byte temporary
+// buffer (around 2 ms per sector), which caps file IO at roughly 150-250 KB/s.
+// The CLR managed heap lives in PSRAM, so every buffer coming from managed code
+// takes that path. Copying through this internal RAM buffer restores multi
+// sector DMA, and the extra memcpy is negligible next to the bus transfer.
+// File system driver calls are serialised by the CLR (SYNC_IO), so a single
+// shared buffer is enough.
+static uint8_t *s_ioBounceBuffer = NULL;
+static const int IO_BOUNCE_BUFFER_SIZE = 16 * 1024;
```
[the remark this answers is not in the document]
dkovyazin: Renamed to `c_ioBounceBufferSize` in 59eb8396, along with the new `c_ioBounceMinTransfer`.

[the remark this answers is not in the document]
dkovyazin: This finding does not apply: on ESP32 the SD volume is served by this driver, not by `FATFS_FS_Driver`.

`FS_MountVolume(mountPoint, 0, "FATFS")` in `nf_sys_io_filesystem_nanoFramework_System_IO_FileSystem_SDCard.cpp` resolves the driver by `FILESYSTEM_DRIVER_INTERFACE::Name`, walking `g_AvailableFSInterfaces` and taking the first match. Both ESP32 interfaces declare the same name: `g_LITTLEFS_FILE_SYSTEM_DriverInterface` ends with `"FATFS", 0,` and so does `g_FATFS_FILE_SYSTEM_DriverInterface` — an upstream copy/paste. In `target_FileSystem.cpp` the littlefs entry comes first, so the lookup always resolves to it, and the SD volume gets `g_LITTLEFS_STREAM_DriverInterface`.

That is also why this driver is a POSIX/VFS wrapper (`fopen`/`fread`/`fwrite`) rather than anything littlefs specific: it serves both the internal partition and the FAT volume of the card through the IDF VFS. The measured improvement on SD transfers comes through exactly this path.

As a side note, `FATFS_FS_Driver::Open()` builds a VFS style path and hands it to `f_open()`, which expects a FatFs path, so that driver looks like it has never actually run on ESP32. Worth a separate issue rather than a change here.

dkovyazin:
Addressed in 59eb8396: verbose comments removed, `IO_BOUNCE_BUFFER_SIZE` renamed to `c_ioBounceBufferSize` per the coding guidelines, and the 512 literal is now `c_ioBounceMinTransfer`. The reasoning about the SDMMC DMA limitation stays in the PR description.

dkovyazin:
The red `ESP32_C3` job (build 65644) failed in `Install IDF` on a `HTTP Error 504` while downloading the xtensa toolchain, so it never reached compilation. For what it is worth on that target specifically: `esp_ptr_external_ram()` is declared unconditionally and returns false when the chip has no PSRAM, so on C3 the bounce path is simply never taken and the code still builds.

dkovyazin:
Correcting myself: not all of build 65644 was infrastructure. The `ESP32_C6_THREAD` and `ESP32_H2_THREAD` jobs got past `Install IDF` and failed to compile, and that one was mine:

```
soc/soc.h:155:42: error: 'SOC_MMU_PAGE_SIZE' was not declared in this scope
  155 | #define SOC_IROM_HIGH   (SOC_IROM_LOW + (SOC_MMU_PAGE_SIZE<<8))
```

`esp_memory_utils.h` has an inline `esp_ptr_in_drom()` that expands `SOC_DROM_HIGH`, and on C6/H2 `soc.h` refers to `SOC_MMU_PAGE_SIZE` without defining it — it comes from `soc/ext_mem_defs.h`. Checked the IDF 5.5.4 headers: only those two chips reference it from `soc.h`, which is exactly the pair that failed. So including that header standalone does not compile there.

Fixed in bb52ec0f by putting the include and the whole bounce path behind `CONFIG_SPIRAM` (with an explicit `sdkconfig.h` include so the symbol is actually defined). C6 and H2 have no external RAM, so on those targets the path was dead code to begin with: `GetIoBounceBuffer()` now takes the pointer and size and returns NULL there, which keeps `Read`/`Write` free of preprocessor branches and lets the compiler drop the bounced path entirely.

The remaining red jobs in that run are still the `Install IDF` HTTP 504 failures.

dkovyazin:
Merged `main` in (10de6927) after #3527 landed — both touch the same spot in `Read`/`Write`, so the branch had started to conflict. Resolution keeps both: the direction change barrier runs first, then the bounce buffer decision, so the `ftell` used for the 4 byte alignment is taken after the stream has been repositioned. The diff against `main` is now only the bounce path.

dkovyazin:
Took the `ftell()` point in 267b3201: the result is stored and a negative value returns `CLR_E_FILE_IO` before the `& 3` mask, in both loops. Without it a failed `ftell` silently became a padding of 3.

The column limit finding does not apply though — the `heap_caps_malloc` line is 118 characters, under the 120 limit, and `awk` over the whole file reports no line above it. `Check_Code_Style` is green on the branch as well.

josesimoes (approved):
LGTM

dkovyazin:
Brought up to date with `main`, which had moved on since the approval: the direction barrier from #3527 and the `access` parameter that #3533 added to `Open()` are both in now. The merge came out clean — `Read()` and `Write()` do the barrier first and then the transfer through the bounce buffer — and nothing in the change itself needed adjusting.

Rebuilt for `ESP32_S3_QUAD` and checked that the bounce path is in the image (the target has external RAM, so it compiles in).

Ready whenever you are.

dkovyazin:
Merged `main` in again (9720e044) — the branch had fallen behind after #3534 landed. Clean merge, the diff against `main` is still the single file with the bounce path and nothing in the change needed adjusting.

Ready to merge whenever the pipeline gets a run.

PR #3529, opened 2026-09-06
Report the SD card capacity in GetSizeInfo

### LINKED ISSUE #1847: DriveInfo.TotalSize is always unknown, GetSizeInfo is stubbed out
**Describe the bug**

`DriveInfo.TotalSize` is always unknown on ESP32. `LITTLEFS_FS_Driver::GetSizeInfo()` is stubbed out: the FAT based calculation is commented out, with a note that `f_getfree()` takes long enough to hit the watchdog, and the function ends with

```c
*totalSize = -1;
*totalFreeSpace = -1;
```

So managed code can never learn the size of a mounted SD card.

**Expected behaviour**

`TotalSize` reports the capacity of the card.

**Proposed fix**

The capacity does not need the FAT walk that the free space does. `sdmmc_card_t` already holds the CSD, read when the card was mounted, so `csd.capacity * csd.sector_size` is available immediately.

The driver also serves the internal flash, which has no CSD, so the capacity has to be reported only for the volume the card is actually mounted on. `Target_System_IO_FileSystem.c` keeps a single `card` for the whole system while there can be several slots, so it needs to record which drive letter that card was mounted under, and clear it on unmount and on a failed mount.

Free space stays unknown for the reason the existing comment gives. PR follows.

dkovyazin:
## Description

- `GetSizeInfo()` reports the capacity of a mounted SD card, taken from the CSD that was read when the card was mounted.
- `Target_System_IO_FileSystem.c` records the drive letter the card is mounted under, and clears it on unmount and on a failed mount.
- Free space stays unknown, for the reason the existing comment gives.

## Motivation and Context

`GetSizeInfo()` is stubbed out: the FAT based calculation is commented out, with a note that `f_getfree()` takes long enough to hit the watchdog, and the function returns -1 for both values. As a result `DriveInfo.TotalSize` is always unknown on ESP32.

The capacity does not need that walk. `sdmmc_card_t` already holds the CSD, so `csd.capacity * csd.sector_size` is available immediately.

The driver also serves the internal flash, which has no CSD, so the capacity may only be reported for the volume the card is actually mounted on. There is a single `card` for the whole system while there can be several slots, hence the recorded drive letter.

- Fixes nanoFramework/Home#1847

## How Has This Been Tested?

- Built for an ESP32-S3 target against current `main`, including the link step.
- Exercised on hardware, where `DriveInfo.TotalSize` for the SD volume reports the card capacity instead of an unknown value, and the internal flash volume still reports unknown.

## Types of changes

## Checklist

### DIFF
--- a/targets/ChibiOS/_FatFs/fatfs_FS_Driver.cpp
+++ b/targets/ChibiOS/_FatFs/fatfs_FS_Driver.cpp
@@ -166,13 +166,11 @@ HRESULT FATFS_FS_Driver::Format(const VOLUME_ID *volume, const char *volumeLabel
 
 HRESULT FATFS_FS_Driver::GetSizeInfo(const VOLUME_ID *volume, int64_t *totalSize, int64_t *totalFreeSpace)
 {
-    (void)totalSize;
-
     // FATFS *fsPtr = &fs;
     // char buffer[3];
     // DWORD freeClusters, freeSectors, totalSectors;
 
-    // FATFS *fs = GetFatFsByVolumeId(volume, false);
+    FATFS *fs = GetFatFsByVolumeId(volume, false);
 
     FileSystemVolume *currentVolume = FileSystemVolumeList::FindVolume(volume->volumeId);
 
@@ -194,8 +192,21 @@ HRESULT FATFS_FS_Driver::GetSizeInfo(const VOLUME_ID *volume, int64_t *totalSize
     //     *totalFreeSpace = (int64_t)freeSectors * FF_MAX_SS;
     // #endif
 
+    // -1 means "unknown" to the caller
     *totalSize = -1;
+
+    // free space would need f_getfree(), see above
     *totalFreeSpace = -1;
+
+    if (fs != NULL)
+    {
+        // capacity of the file system, from the mounted FATFS object, without walking the FAT
+#if FF_MAX_SS != FF_MIN_SS
+        *totalSize = (int64_t)(fs->n_fatent - 2) * fs->csize * fs->ssize;
+#else
+        *totalSize = (int64_t)(fs->n_fatent - 2) * fs->csize * FF_MAX_SS;
+#endif
+    }
 
     return S_OK;
 }
--- a/targets/ESP32/_common/Target_System_IO_FileSystem.c
+++ b/targets/ESP32/_common/Target_System_IO_FileSystem.c
@@ -45,11 +45,7 @@ static const char *TAG = "SDCard";
 
 sdmmc_card_t *card;
 
-// Drive letter the card in `card` is mounted under, 0 when it is not mounted.
-// There is a single `card` for the whole system while there can be several
-// slots, so without this binding a consumer could not tell its own volume from
-// another one and would report the characteristics of the first card for the
-// second slot.
+// drive letter the card above is mounted under, 0 when it is not mounted
 char cardDriveLetter;
 
 //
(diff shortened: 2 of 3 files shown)


josesimoes (changes requested):
Same here: verbose comments. Reasoning belongs in the PR comments, unless it's not obvious and it deserves a small note on the source code.

josesimoes:
@coderabbitai review

josesimoes:
@dkovyazin any chance you can add this fix to ChibiOS targets too? The code is quite similiar (if not the same)...
Please take a look at targets\ChibiOS\_FatFs\fatfs_FS_Driver.cpp::GetSizeInfo()

dkovyazin:
Shortened in 54bfaf37 — one short line where it is not obvious, the rest is in the PR description.

dkovyazin:
Added in 5b6da208, though through a different source than the ESP32 side.

There is no CSD to read on ChibiOS, but `GetSizeInfo()` there already had the answer next to it: `GetFatFsByVolumeId()` gives the mounted `FATFS` object, and `(n_fatent - 2) * csize * ssize` is the capacity of the file system. That is the part of the commented-out block right above it that never needed `f_getfree()` — only the free space does, which is why it stays unknown. The multiplication starts from an `int64_t` so a large volume cannot overflow the intermediate `DWORD`.

So `totalSize` is the file system capacity on ChibiOS and the physical card capacity on ESP32. Happy to align them if you prefer one over the other — on ESP32 the driver is a VFS wrapper with no `FATFS` object at hand, which is why it goes through the card CSD there.

Caveat worth stating plainly: the ESP32 side was measured on hardware, this one was not — I have no ChibiOS board here, so it has only been reviewed and will only be covered by the build.

josesimoes:
@codereabbitai review

dkovyazin:
@coderabbitai review

(the handle above had a typo, so the bot did not pick it up)

dkovyazin:
@coderabbitai this is the same finding you raised on #3528 and withdrew after verifying it there. The registry does not resolve `"FATFS"` to `FATFS_FS_Driver` on ESP32.

`FS_MountVolume()` in `src/PAL/FileSystem/nanoPAL_FileSystem.cpp` walks `g_AvailableFSInterfaces`, compares `fsDriver->Name` and `break`s on the **first** match. On ESP32 both interfaces declare the same name — `g_LITTLEFS_FILE_SYSTEM_DriverInterface` ends with `"FATFS", 0,` (littlefs_FS_Driver.cpp:1153) and so does `g_FATFS_FILE_SYSTEM_DriverInterface` (fatfs_FS_Driver.cpp:1022), an upstream copy/paste. In every `targets/ESP32/*/target_FileSystem.cpp` the littlefs entry is listed first, so `MountNative___VOID()` calling `FS_MountVolume(mountPoint, 0, "FATFS")` binds the SD volume to `g_LITTLEFS_STREAM_DriverInterface` and `LITTLEFS_FS_Driver`.

That is also why this driver is written against `fopen`/`fread`/`fwrite` rather than the littlefs API: it is a VFS wrapper serving both the internal partition and the FAT volume of the card.

Moving the code to `FATFS_FS_Driver::GetSizeInfo()` would put it in a driver that is never reached on ESP32. Its `Open()` even builds a VFS style path and passes it to `f_open()`, which expects a FatFs path — worth a separate issue, not a change here.

josesimoes (approved):
LGTM

PR #3532, opened 2026-09-19
Fix/event cache initialize

Wielebny666:
## Description

- Replaced the hardcoded `INLINE_SIZE_BYTES` constant in `CLR_RT_RuntimeMemory.cpp` with `sizeof(CLR_RT_InlineBuffer)`, so the scratch buffer is sized from the actual type instead of a layout that `CLR_RT_InlineFrame` outgrew long ago.
- Fixed `g_scratchInlineBuffer` being allocated at less than half the required size (36 B vs 76 B per entry on 32-bit ARM, 68 B vs 144 B on 64-bit), which made `CLR_RT_EventCache::EventCache_Initialize()` write its free list well past the end of the array and corrupt adjacent `.bss` on every CLR startup.
- Added a `static_assert` that keeps `g_scratchInlineBuffer` large enough for `PLATFORM_DEPENDENT_INLINE_BUFFER_SIZE` inline frames, so any future change to the frame layout fails at compile time instead of silently corrupting memory.
- Added the `nanoCLR_Runtime.h` include required to make `CLR_RT_InlineBuffer` a complete type in this translation unit.

## Motivation and Context

- Fixes

## How Has This Been Tested?

ASan detected the error in a 32-bit build of the POSIX target: upon CLR startup, it reported an out-of-bounds write past the end of the global buffer used for inline frames. The buffer size is calculated based on a constant of 36 bytes per element, whereas the actual element size is 76 bytes; consequently, the buffer is less than half the required size. The constant dates back to the project's inception and was not updated when new fields were added to the structure. This discovery was accidental, as I was merely verifying that ASan worked on the new target. I did not test this on other targets or actual hardware, so for those platforms, the conclusion is based on code analysis rather than empirical measurement.
## Types of changes

## Checklist


Wielebny666:
@dotnet-policy-service agree

josesimoes:
@coderabbitai review

Wielebny666:
Hi,
@josesimoes For the TI_CC1352R1_LAUNCHXL target, the NF_CLR_NO_IL_INLINE flag needs to be set, but I don't think this MR is the right place for it.

PR #3534, opened 2026-09-21
Enter the ROM download mode on EnterProprietaryBooter

dkovyazin:
## Description

- `RequestToLaunchProprietaryBootloader()` becomes weak, as its neighbour `LaunchProprietaryBootloader()` already is.
- The ESP32 target implements it by setting FORCE_DOWNLOAD_BOOT, which takes the ROM into download mode on the reset that the caller performs next.
- Implemented for every supported series: RTC_CNTL on C3, S2 and S3, LP_AON on C5, C6, C61 and H2, LP_SYS on P4. The classic ESP32 has no such bit and keeps the weak default, which answers "not supported" as it does today.

## Motivation and Context

`Monitor_Reboot` carries `c_EnterProprietaryBooter` and the CLR honours it, but on ESP32 the request always failed: the generic implementation only answers on targets that have nanoBooter. These chips do have a bootloader of their own, the ROM download mode, and reaching it was a manual job. On a board that talks over the chip's own USB, esptool cannot drive EN and IO0, so the strapping pin has to be held by hand for every reflash, including on boards that are running nanoFramework and are perfectly able to answer a command.

FORCE_DOWNLOAD_BOOT lives in the always-on domain: it survives the software reset that follows and takes the ROM into download mode regardless of the strapping pin. A power cycle clears it. The same write is what ESP-IDF itself does in `esp_system/port/usb_console.c`.

Both callers reach this through `EnterProprietaryBooter` already: the debugger command in `Debugger.cpp` and the managed `Power.RebootDevice(RebootOption.EnterProprietaryBooter)` through `Execution.cpp`.

As discussed in the issue, the follow-ups live outside this repository: nanoff sends the flag before flashing an ESP32 that answers the wire protocol, and the Device Explorer reboot drop-down needs no change at all, since it already maps to `RebootOptions.EnterProprietaryBooter` and enables itself from `HasProprietaryBooter`, which the ESP32 targets already report.

- Adresses nanoFramework/Home#1849

## How Has This Been Tested?

- On an ESP32-S3 with native USB (custom target, RTM build), driven by a build of nanoff that sends the flag: the board takes the command, reboots into the ROM and is flashed without the BOOT button being touched. Disassembly of the linked image confirms the write lands on `RTC_CNTL_OPTION1_REG`.
- Built for `ESP32_C6_THREAD`, `ESP32_C5_THREAD` and `ESP32_P4` and checked the generated code: a read-modify-write of `LP_AON_SYS_CFG_REG` setting bit 30 on the C6, the two bit field set to `01` (download boot over UART and USB) on the C5, and bit 2 of `LP_SYSTEM_REG_SYS_CTRL_REG` on the P4. No hardware here for those parts, so that is as far as it goes.
- Register addresses and masks come from the ESP-IDF v5.5.4 headers. The three that esptool also carries (S2, S3, P4) match its values.

Two things worth recording:

- `CLRStartup` also calls this function under `RevertToBooterOnFault`. It does not fire on ESP32, since every `app_main` zeroes `CLR_SETTINGS`, but a target that sets it would land in the ROM instead.
- On S2 and S3 with the USB-OTG console, IDF additionally sets the USB persist flags before rebooting into the ROM. This change does not, so such a board re-enumerates on its way in. It is not a problem on the USB-Serial/JTAG boards this was tested on, and the host tools handle the re-enumeration.

## Types of changes

## Checklist


josesimoes (approved):
LGTM

